PR #14, opened 2020-03-29
podman_image: only set changed=true if there is a new image

larsks:
with force: true, podman_image would always report changed regardless
of whether or not the pull or build operation resulted in a new image.
With this commit, check the image digest before and after and only
reported changed if the digest is different.

I'm relocating this from https://github.com/ansible/ansible/pull/68542

### DIFF
--- a/plugins/modules/podman_image.py
+++ b/plugins/modules/podman_image.py
@@ -424,19 +424,26 @@ class PodmanImageManager(object):
     def present(self):
         image = self.find_image()
 
+        if image:
+            digest_before = image[0].get('Digest', image[0].get('digest'))
+        else:
+            digest_before = None
+
         if not image or self.force:
             if self.path:
                 # Build the image
                 self.results['actions'].append('Built image {image_name} from {path}'.format(image_name=self.image_name, path=self.path))
-                self.results['changed'] = True
                 if not self.module.check_mode:
                     self.results['image'] = self.build_image()
             else:
                 # Pull the image
                 self.results['actions'].append('Pulled image {image_name}'.format(image_name=self.image_name))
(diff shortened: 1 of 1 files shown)


plugins/modules/podman_image.py
```diff
@@ -424,20 +424,27 @@ def _get_id_from_output(self, lines, startswith=None, contains=None, split_on='
     def present(self):
         image = self.find_image()
 
+        if image:
+            digest_before = image[0]['Digest']
```
sshnaidm: Unfortunately podman inspect info is not consistent along versions, so in 1.8 it's everything lowercase - `"digest"`
I'd suggest to force lower case in inspection function `self.find_image()` so that we can use only `['digest']`
Or something like: 
```python
digest_before = image[0].get('Digest', image[0].get('digest'))
```

plugins/modules/podman_image.py
```diff
@@ -424,20 +424,27 @@ def _get_id_from_output(self, lines, startswith=None, contains=None, split_on='
...
                     self.results['image'] = self.build_image()
             else:
                 # Pull the image
                 self.results['actions'].append('Pulled image {image_name}'.format(image_name=self.image_name))
-                self.results['changed'] = True
                 if not self.module.check_mode:
                     self.results['image'] = self.pull_image()
 
+            image = self.find_image()
+            digest_after = image[0]['Digest']
```
sshnaidm: same as above

plugins/modules/podman_image.py
larsks: Yes, I had noticed that myself while trying to use this earlier today! What a pain. I'll fix it up.

PR #15, opened 2020-03-30
Github workflow

sshnaidm:
Add sanity tests for current modules, except of podman_container which requires specific changes.
Add building collection jobs.

### DIFF
--- a/.github/workflows/collection-continuous-integration.yml
+++ b/.github/workflows/collection-continuous-integration.yml
@@ -1,0 +1,158 @@
+name: Collection test suite
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: 3 0 * * *  # Run daily at 0:03 UTC
+
+jobs:
+  build-collection-artifact:
+    name: Build collection
+    runs-on: ${{ matrix.runner-os }}
+    strategy:
+      matrix:
+        runner-os:
+          - ubuntu-latest
+          - ubuntu-16.04
+        ansible-version:
+          - ansible==2.9.5
+          - git+https://github.com/ansible/ansible.git@devel
+        runner-python-version:
+          - 2.7
+          - 3.7
+        exclude:
+          - runner-os: ubuntu-latest
+            runner-python-version: 2.7
+
+    steps:
+      - name: Check out ${{ github.repository }} on disk
+        uses: actions/checkout@master
+      - name: Set up Python ${{ matrix.runner-python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.runner-python-version }}
+      - name: Set up pip cache
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ hashFiles('tests/sanity/requirements.txt') }}-${{ hashFiles('tests/unit/requirements.txt') }}
+          restore-keys: |
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
+      - name: Install Ansible ${{ matrix.ansible-version }}
+        run: >-
+          python -m pip install --user ${{ matrix.ansible-version }}
+      - name: Build a collection tarball
+        run: >-
+          ~/.local/bin/ansible-galaxy collection build --output-path
+          "${GITHUB_WORKSPACE}/.cache/collection-tarballs"
+      - name: Store migrated collection artifacts
+        uses: actions/upload-artifact@v1
+        with:
+          name: collection
+          path: .cache/collection-tarballs
+
+  sanity-test-collection-via-containers:
+    name: Sanity in container via Python ${{ matrix.python-version }}
+    needs:
+      - build-collection-artifact
+    runs-on: ${{ matrix.runner-os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        runner-os:
+          - ubuntu-latest
+        runner-python-version:
+          - 3.7
+        ansible-version:
+          - ansible==2.9.5
+          - git+https://github.com/ansible/ansible.git@devel
+        python-version:
+          - 2.7
+          - 3.7
+
+    steps:
+      - name: Set up Python ${{ matrix.runner-python-version }}
+        uses: actions/setup-python@v1
+        with:
+          python-version: ${{ matrix.runner-python-version }}
+      - name: Set up pip cache
+        uses: actions/cache@v1
+        with:
+          path: ~/.cache/pip
+          key: ${{ runner.os }}-pip-${{ github.ref }}-sanity-containers
+          restore-keys: |
+            ${{ runner.os }}-pip-
+            ${{ runner.os }}-
+      - name: Install Ansible ${{ matrix.ansible-version }}
+        run: >-
+          python -m pip install --user ${{ matrix.ansible-version }}
+      - name: Download migrated collection artifacts
+        uses: actions/download-artifact@v1
+        with:
+          name: collection
+          path: .cache/collection-tarballs
+      - name: Install the collection tarball
+        run: >-
+          ~/.local/bin/ansible-galaxy collection install .cache/collection-tarballs/*.tar.gz
+      - name: Run collection sanity tests
+        run: >-
+          ~/.local/bin/ansible-test sanity --color --requirements --docker --python "${{ matrix.python-version }}" -vvv
+        working-directory: >-
+          /home/runner/.ansible/collections/ansible_collections/containers/podman
+
+  unit-test-collection-via-containers:
+    name: Units in container ${{ matrix.container-image }}
+    needs:
+      - build-collection-artifact
+    runs-on: ${{ matrix.runner-os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        runner-os:
+          - ubuntu-latest
+        runner-python-version:
+          - 3.7
+        ansible-version:
+          - git+https://github.com/ansible/ansible.git@devel
+        container-image:
+          - fedora31
+          #- ubuntu1804
+          - centos8
+          #- opensuse15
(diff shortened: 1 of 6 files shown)


.github/workflows/collection-continuous-integration.yml
```diff
@@ -0,0 +1,258 @@
...
+    steps:
+    - name: Set up Python ${{ matrix.runner-python-version }}
+      uses: actions/setup-python@v1
+      with:
+        python-version: ${{ matrix.runner-python-version }}
+    - name: Set up pip cache
+      uses: actions/cache@v1
+      with:
+        path: ~/.cache/pip
+        key: ${{ runner.os }}-pip-${{ github.ref }}-sanity-containers
```
webknjaz: For best results, you should also bake-in the deps files hashes

.github/workflows/collection-continuous-integration.yml
```diff
@@ -0,0 +1,258 @@
...
+        restore-keys: |
+          ${{ runner.os }}-pip-
+          ${{ runner.os }}-
+    - name: Install Ansible ${{ matrix.ansible-version }}
+      run: >-
+        python -m pip install --user ${{ matrix.ansible-version }}
+    - name: Download migrated collection artifacts
+      uses: actions/download-artifact@v1
+      with:
+        name: >-
```
webknjaz: This field is very short, one word does not need multilining

.github/workflows/collection-continuous-integration.yml
```diff
@@ -0,0 +1,258 @@
...
+          collection
+        path: .cache/collection-tarballs
+    - name: Install the collection tarball
+      run: >-
+        ~/.local/bin/ansible-galaxy collection install .cache/collection-tarballs/*.tar.gz
+    - name: Run collection sanity tests
+      run: >-
+        ~/.local/bin/ansible-test sanity --color --requirements --docker --python "${{ matrix.python-version }}" -vvv
```
webknjaz: I'm curious if it'd be faster w/o `--docker`. Also, `--python` arg probably isn't necessary.

webknjaz: Oh, nevermind. I see what you're doing. OTOH you could install all the necessary Python interpreters on the VM directly... The benefit of using the pre-baked container is that it has more Python versions than GH Actions support. So you may want to reconsider.

.github/workflows/collection-continuous-integration.yml
```diff
@@ -0,0 +1,258 @@
...
+  - cron: 3 0 * * *  # Run daily at 0:03 UTC
+
+jobs:
+  build-collection-artifact:
+    name: Build collection
+    runs-on: ${{ matrix.runner-os }}
+    strategy:
+      matrix:
+        runner-os:
+          - ubuntu-latest
```
webknjaz: Your sequence indentations are inconsistent across this file. I recommend that you choose one and stick with it. Personally I don't like indented sequences. It is common to document such preference by committing a YAML lint config, e.g. https://github.com/ansible-community/collection_migration/blob/master/.yamllint#L2

.github/workflows/collection-continuous-integration.yml
sshnaidm: yeah, good idea, will do overall optimization in followups

.github/workflows/collection-continuous-integration.yml
sshnaidm: migrate.py does it :)
fixed

.github/workflows/collection-continuous-integration.yml
sshnaidm: Yes, it was w/o indentation, I've added it intentionally. For me it's difficult to parse the file w/o indentation.

sshnaidm: fixed

.github/workflows/collection-continuous-integration.yml
sshnaidm: thanks, I'll try to get to it when optimizing jobs.

.github/workflows/collection-continuous-integration.yml
webknjaz: I know, I authored that for migrate.py. But I only did it in places where I wanted it to fit 80-char line-length (because short lines are better readable. If you look at https://github.com/ansible-community/collection_migration/blob/master/.github/workflows/collection-migration-tests.yml, you'll see that for short names it's not used.
It's also used in places with templating — you cannot split those lines because YAML will render an extra space in the middle.

Another good use case is strings starting with `{` — normally you'd have to quote them but not if `>` or `|` is used.

.github/workflows/collection-continuous-integration.yml
```diff
@@ -0,0 +1,158 @@
...
+        uses: actions/download-artifact@v1
+        with:
+          name: collection
+          path: .cache/collection-tarballs
+      - name: Install the collection tarball
+        run: >-
+          ~/.local/bin/ansible-galaxy collection install .cache/collection-tarballs/*.tar.gz
+      - name: Run collection unit tests
+        run: |
+          [[ ! -d 'tests/unit' ]] && echo This collection does not have unit tests. Skipping... || \
```
webknjaz: This conditional is unnecessary. It was only needed in `migrate.py` because sometimes we had resulting collections with no tests. If you know that you have tests, remove this or it may start silently start skipping testing at some point in the future.

.github/workflows/collection-continuous-integration.yml
```diff
@@ -0,0 +1,158 @@
...
+        run: >-
+          python -m pip install --user ${{ matrix.ansible-version }}
+      - name: Build a collection tarball
+        run: >-
+          ~/.local/bin/ansible-galaxy collection build --output-path
+          "${GITHUB_WORKSPACE}/.cache/collection-tarballs"
+      - name: Store migrated collection artifacts
+        uses: actions/upload-artifact@v1
+        with:
+          name: collection
```
webknjaz: @sshnaidm

Because this job is a part of the matrix, you need to have different names for artifacts or they'll probably overwrite each other. So there's no guarantee which of the jobs produces the artifact that the next one will download.

Also, I'm not sure if it makes sense to build all of the permutations. Maybe it's okay to build just one but test it under different envs? After all, you'll upload only one tarball to Galaxy and it's what the end-users will receive, not all possible variants.

.github/workflows/collection-continuous-integration.yml
```diff
@@ -0,0 +1,158 @@
...
+    needs:
+      - build-collection-artifact
+    runs-on: ${{ matrix.runner-os }}
+    strategy:
+      fail-fast: false
+      matrix:
+        runner-os:
+          - ubuntu-latest
+        runner-python-version:
+          - 3.7
```
webknjaz: I'd use 3.8 already.

.github/workflows/collection-continuous-integration.yml
```diff
@@ -0,0 +1,158 @@
...
+
+on:
+  push:
+  pull_request:
+  schedule:
+    - cron: 3 0 * * *  # Run daily at 0:03 UTC
+
+jobs:
+  build-collection-artifact:
+    name: Build collection
```
webknjaz: FYI these `name` fields show up in many weird places on the GH UI and most of the time the visual area is limited.
For the best results, invent as short names as possible. That's why I even use single-emoji in some places.

.github/workflows/collection-continuous-integration.yml
sshnaidm: Yeah, but actually we don't have unittests right now. But very want to have them. :) So maybe we'll leave it until we have tests and then will remove this condition.

.github/workflows/collection-continuous-integration.yml
sshnaidm: Agree, I'll make only one job building, it shouldn't matter which platform is used.

.github/workflows/collection-continuous-integration.yml
sshnaidm: We still use 3.7 widely, although I think it's worth to add 3.8 in some point.

.github/workflows/collection-continuous-integration.yml
sshnaidm: I noticed the same, trying to understand which job I'm tracking right now, it's really annoying I can't see the name. Hopefully GH will fix it someday.

plugins/modules/podman_container.py
```diff
@@ -1766,11 +1767,13 @@ def stopped(self):
 
     def get_info(self):
         """Inspect container and gather info about it."""
+        # pylint: disable=unused-variable
         rc, out, err = self.module.run_command(
```
webknjaz: Instead of adding ignore comments, it's better to actually fix problems.
In this case it's as easy as adding an underscore to unused vars. Like `err` -> `_err`.

.github/workflows/collection-continuous-integration.yml
webknjaz: In this case, I'm pretty sure you could just add something like `if: false` to the step definition. This way, it'd visually show up as skipped, not succeeded in the CI.

webknjaz: Moreover, you could add another step that will always fail if unit tests exist, so it'd prompt you to remove `if: false` once somebody adds actual tests.

plugins/modules/podman_container.py
sshnaidm: Well, it won't work "out of box", you need to pass `--dummy-variables-rgx=^_.*` to pylint for that.

webknjaz: Hm... Well, it's set up in a typical pylint default config. I guess I forgot that Ansible's preset may be different.

PR #19, opened 2020-04-09
Add user flags before container id in podman exec

Akasurde:
##### SUMMARY

When user provides an ansible_ssh_user, podman connection
plugin includes this values as `--user` flag. This patch
fixes the location of this flag according to podman exec command help.

Fixes: ansible/ansible#65220

Signed-off-by: Abhijeet Kasurde <akasurde@redhat.com>

##### COMPONENT NAME
plugins/connection/podman.py


Akasurde:
cc @rhatdan

sshnaidm:
@Akasurde thanks, let me test it locally while no tests in CI

plugins/connection/podman.py
```diff
@@ -148,10 +153,11 @@ def exec_command(self, cmd, in_data=None, sudoable=False):
 
         # shlex.split has a bug with text strings on Python-2.6 and can only handle text strings on Python-3
         cmd_args_list = shlex.split(to_native(cmd, errors='surrogate_or_strict'))
+        exec_args_list = ["exec"]
         if self.user:
-            cmd_args_list += ["--user", self.user]
+            exec_args_list.extend(("--user", self.user))
```
rhatdan, line 158: I would think it would be better to prepend to cmd_args_list.

Basically the OPTIONS have to preceed the Container ID>

Akasurde: I had the same idea before but the container name comes before `cmd_args_list` and podman exec does not like flags after container name.

```
podman exec [flags] CONTAINER [COMMAND...
```

rhatdan: Ok I misunderstood the what this was doing.

plugins/connection/podman.py
```diff
@@ -108,6 +108,11 @@ def _podman(self, cmd, cmd_args=None, in_data=None, use_container_id=True):
                 to_native(
                     self.get_option('podman_extra_args'),
                     errors='surrogate_or_strict'))
+        if isinstance(cmd, str):
+            local_cmd.append(cmd)
+        else:
+            local_cmd.extend(cmd)
+
         local_cmd.append(cmd)
```
sshnaidm, line 116: duplicate of line 112?

plugins/connection/podman.py
sshnaidm, line 158: Something is wrong here, from the test:
```
TASK [Gathering Facts] ********************************************************************************************************************************************************************************************
task path: /tmp/test2.yml:2
<podman-container> RUN [b'/usr/bin/podman', b'mount', b'mount', b'podman-container']
Failed to mount container podman-container: b'Error: cannot mount using driver overlay in rootless mode'
<podman-container> RUN [b'/usr/bin/podman', b'exec', b'--user', b'nobody', b"['exec', '--user', 'nobody']", b'podman-container', b'/bin/sh', b'-c', b'echo ~nobody && sleep 0']
<podman-container> RUN [b'/usr/bin/podman', b'exec', b'--user', b'nobody', b"['exec', '--user', 'nobody']", b'podman-container', b'/bin/sh', b'-c', b'echo "`pwd`" && sleep 0']
<podman-container> RUN [b'/usr/bin/podman', b'exec', b'--user', b'nobody', b"['exec', '--user', 'nobody']", b'podman-container', b'/bin/sh', b'-c', b'( umask 77 && mkdir -p "` echo ~/.ansible/tmp `"&& mkdir ~/.a
nsible/tmp/ansible-tmp-1587249105.6848736-22218-208278136013092 && echo ansible-tmp-1587249105.6848736-22218-208278136013092="` echo ~/.ansible/tmp/ansible-tmp-1587249105.6848736-22218-208278136013092 `" ) && sl
eep 0']
fatal: [podman-container]: UNREACHABLE! => {
    "changed": false,
    "msg": "Failed to create temporary directory.In some cases, you may have been able to authenticate and did not have permissions on the target directory. Consider changing the remote tmp path in ansible.cfg t
o a path rooted in \"/tmp\", for more error information use -vvv. Failed command was: ( umask 77 && mkdir -p \"` echo ~/.ansible/tmp `\"&& mkdir ~/.ansible/tmp/ansible-tmp-1587249105.6848736-22218-20827813601309
2 && echo ansible-tmp-1587249105.6848736-22218-208278136013092=\"` echo ~/.ansible/tmp/ansible-tmp-1587249105.6848736-22218-208278136013092 `\" ), exited with result 125, stderr output: Error: no container with 
name or ID ['exec', '--user', 'nobody'] found: no such container\n",
    "unreachable": true
}

```

for reproducing:
```bash
podman run -d --name podman-container python:3-alpine sleep 1d
ansible-playbook -vvvv test2.yml -i podman-container, -u nobody
```
when test2.yml is:
```yaml
---
- hosts: podman-container
  connection: containers.podman.podman
  tasks:

    - name: Print who am I
      shell: whoami
```

**please be aware to use collections connection, not the build one in 2.9**

sshnaidm: With new patch works like a charm, thanks. Will merge it as we'll have CI working in next few days.

sshnaidm:
need to update the test https://github.com/containers/ansible-podman-collections/blob/3d1c38c3c2d38b86ed1d746c638623e0af999180/tests/integration/targets/connection_podman/runme.sh#L18

plugins/connection/podman.py
```diff
@@ -149,10 +153,11 @@ def exec_command(self, cmd, in_data=None, sudoable=False):
 
         # shlex.split has a bug with text strings on Python-2.6 and can only handle text strings on Python-3
         cmd_args_list = shlex.split(to_native(cmd, errors='surrogate_or_strict'))
+        exec_args_list = ["exec"]
         if self.user:
-            cmd_args_list += ["--user", self.user]
+            exec_args_list.extend(("--user", self.user))
 
-        rc, stdout, stderr = self._podman("exec", cmd_args_list, in_data)
+        rc, stdout, stderr = self._podman(exec_args_list, cmd_args_list, in_data)
```
ktdreyer, line 160: I think this is a problem that we're passing a list here instead of a string. `_podman()` expects a string.

ktdreyer: My bad, I was looking at the Ansible v2.9 code!

ktdreyer: I looked at this again, and this change actually applies to both Ansible 2.9 and Ansible 2.10+. There are some small differences in `_podman()` between 2.9 and 2.10, but in particular, this commit changes the type of the `cmd` parameter that `_podman()` will accept. I've updated the docstring to make this clear, at https://github.com/containers/ansible-podman-collections/pull/79

ktdreyer: I've cherry-picked this for Ansible 2.9 here: https://github.com/ansible/ansible/pull/70541

sshnaidm: @ktdreyer you might want to cherry-pick buildah change too: https://github.com/containers/ansible-podman-collections/commit/1c3d2456b7a5270ff264407ee927074e9ad8961a

ktdreyer: If it's ok with you, I will leave that up to someone else, because I don't personally use the buildah plugin. I just need this change for Molecule with Ansible 2.9, because I have several playbooks that use "become".

sshnaidm: @ktdreyer  Sure, it's not something urgent. Can you please point me in your time to example of usage podman connection plugin with molecule? I'm curios about the use cases. 
Thanks

ktdreyer: Sure. I have [a repository with playbooks](https://github.com/ktdreyer/koji-playbooks) that set up an application for me ([Koji](https://pagure.io/koji)). I use Molecule to set up and test Koji within a podman container.

 I'm using `become: yes` and `become_method: sudo` in my [top-level Molecule playbook](https://github.com/ktdreyer/koji-playbooks/blob/master/molecule/centos8/converge.yml). This means I run all tasks as `root` by default, and then I have several Kerberos-related tasks that I run as as the unprivileged user with `become: no`.

I found that Ansible would always run everything as root, even with `become: no`, because the podman connection plugin was running `podman exec` with the `--user` flag in the incorrect position. With this patch, Ansible 2.9 correctly runs the tasks as the unprivileged user.

PR #26, opened 2020-04-25
buildah/connection add support of specific user

Raoul555:
Add possibility to use a specific user, either defined by --user argument on
command line, or ansible_user, to connect to containers.

It is inspired from
https://github.com/containers/ansible-podman-collections/pull/19 the equivalent
for podman connection.

It was laso required to change the method to put file, from a mount mechanism,
to the buildah copy function, to allow to set correctly the permissions of
pushed files.

Fixes:  containers/ansible-podman-collections#25

### DIFF
--- a/plugins/connection/buildah.py
+++ b/plugins/connection/buildah.py
@@ -78,13 +78,14 @@ class Connection(ConnectionBase):
     def _set_user(self):
         self._buildah(b"config", [b"--user=" + to_bytes(self.user, errors='surrogate_or_strict')])
 
-    def _buildah(self, cmd, cmd_args=None, in_data=None):
+    def _buildah(self, cmd, cmd_args=None, in_data=None, outfile_stdout=None):
         """
         run buildah executable
 
         :param cmd: buildah's command to execute (str)
         :param cmd_args: list of arguments to pass to the command (list of str/bytes)
         :param in_data: data passed to buildah's stdin
+        :param outfile_stdout: file for writing STDOUT to
         :return: return code, stdout, stderr
         """
         local_cmd = ['buildah', cmd, '--', self._container_id]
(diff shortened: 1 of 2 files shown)


sshnaidm:
Thanks, going to test it. There are some sanity failures:
https://github.com/containers/ansible-podman-collections/pull/26/checks?check_run_id=618363814#step:7:2841

Please don't shorten lines, we have max line length 119 and it distracts from real functional changes. The patch is better to change only one issue.

plugins/connection/buildah.py
```diff
@@ -129,18 +144,23 @@ def put_file(self, in_path, out_path):
...
-        shutil.copyfile(
-            to_bytes(in_path, errors='surrogate_or_strict'),
-            to_bytes(real_out_path, errors='surrogate_or_strict')
-        )
-        # alternatively, this can be implemented using `buildah copy`:
-        # rc, stdout, stderr = self._buildah(
-        #     "copy",
-        #     [to_bytes(in_path, errors='surrogate_or_strict'),
-        #      to_bytes(out_path, errors='surrogate_or_strict')]
+        # if put_file use buildah mount, we won't able to set correctly
```
sshnaidm, line 147: So it won't work with mounts at all? Then we should either to remove mounts completely and use `copy` only like in #28 
Or to find a way to use mounts with user

Raoul555: I kept mount in this version, as it is still used for fetching files.

sshnaidm: Can you please rebase on the last version? Thanks

sshnaidm:
@Raoul555 are you planing to continue with this task?

Raoul555:
Hi @sshnaidm 

I've update the pull request, but it fails on testauto.

Testauto are run with ANSIBLE_REMOTE_USER = 1000, but this user seems not existing in the container.

Do you know why this test settup?

sshnaidm:
@Raoul555 there is actually user "nobody" with 1000, but it doesn't have login.
Change the container to something with user, let's see if it helps: https://github.com/containers/ansible-podman-collections/blob/804b60e4fd4066546d0bcdd7f98db39bab1ffead/ci/run_connection_test.sh#L15
instead of `python:2`

PR #27, opened 2020-05-01
Set up testing for podman and buildah connection

sshnaidm:
Set up CI and testing for podman and buildah connection plugins


sshnaidm:
@Akasurde please rebase your #19 after this PR #27 merges, let's see how CI works.

PR #28, opened 2020-05-02
Buildah connection rootless

sshnaidm:
Support rootless containers with buildah connection
Fixes https://github.com/ansible/ansible/issues/63316


sshnaidm:
@TomasTomecek can you please take a look? I'd like to have it as a temporary solution for buildah rootless, until I make it work with `unshare`, if it's possible. Currently `unshare mount` doesn't work for me well.

PR #42, opened 2020-05-11
buildah/connection add support of specific user

Raoul555:
Add possibility to use a specific user, either defined by --user argument on
command line, or ansible_user, to connect to containers.

It is inspired from
https://github.com/containers/ansible-podman-collections/pull/19 the equivalent
for podman connection.

It was laso required to change the method to put file, from a mount mechanism,
to the buildah copy function, to allow to set correctly the permissions of
pushed files.

Fixes:  containers/ansible-podman-collections#25

### DIFF
--- a/plugins/connection/buildah.py
+++ b/plugins/connection/buildah.py
@@ -88,10 +88,28 @@ class Connection(ConnectionBase):
         :param outfile_stdout: file for writing STDOUT to
         :return: return code, stdout, stderr
         """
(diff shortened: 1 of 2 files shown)


sshnaidm:
@Raoul555 thanks, going to test it today.

sshnaidm:
@TomasTomecek do you want to take a look maybe?

sshnaidm:
There is a small problem with buildah connection on py2, I'll solve it in followups.

PR #109, opened 2020-09-14
podman_image: reuse existing results in present()

cfelder:
This patch also helps when pulling images using a "non-canonical name", e.g.

docker.io/mariadb:10.1 which points to docker.io/library/mariadb:10.1.

``digest_before`` will always be``None`` but at least ``digest_after`` can be calculated using
``digest_after = image[0].get('Digest', image[0].get('digest'))``

whereas ``find_image()`` would still return ``None``


sshnaidm:
@cfelder thanks! Can you think about testing this?

sshnaidm:
@cfelder I'd be very appreciated if you elaborate a bit more in commit message what is this change for and the use case. We can think about tests in followups.

cfelder:
@sshnaidm Sorry for my delayed response. I am currently on vacation and will get back to this later. I am also interested in providing tests for this. This will probably shed more light on this.

cfelder:
example:
```
- name: running single containe
    containers.podman.podman_image:
      name: "docker.io/nextcloud:stable-fpm"
```

The image name docker.io/nextcloud:stable-fpm normalises to docker.io/**library**/nextcloud:stable-fpm

https://github.com/containers/ansible-podman-collections/blob/e5b38f097b407ecdf25127421d63c059d100c26b/plugins/modules/podman_image.py#L465

this pull resolves to docker.io/**library**/nextcloud:stable-fpm

https://github.com/containers/ansible-podman-collections/blob/e5b38f097b407ecdf25127421d63c059d100c26b/plugins/modules/podman_image.py#L467-L468

Without storing the result on ``image`` there will be a follow up call to ``find_image()`` that still makes a lookup using

``image_name = "docker.io/nextcloud:stable-fpm"`` which doesn't resolve on ``podman image ls`` because the image has been stored using the normalised url containing **library/**

https://github.com/containers/ansible-podman-collections/blob/e5b38f097b407ecdf25127421d63c059d100c26b/plugins/modules/podman_image.py#L493-L496


With this patch ``digest_after`` can be calculated and ``result['changed']`` will report ``True`` whereas the previous version always reports ``False`` (``digest_before`` and ``digest_after`` are both ``None``) although a new image may has been pulled

PR #111, opened 2020-09-28
Fix init: option it's boolean not string

ruatag:
If init: true/yes is set it passed as string argument and produces error.
Default init path is predefined as /dev/init and automically injected from catatonit package https://github.com/openSUSE/catatonit
Custom can be set with --init-path=


sshnaidm (approved):

sshnaidm:
@ruatag thanks! If you can also add test for this option, would be great.

Issue #113, opened 2020-10-05
config diff calculation is not able to handler upper / lower case correctly


HolgerHees:
In all *Diff: classes where we test the configuration for changes we dump the current config to json, convert it to lower case and load it again into an dict object.

e.g: podman_container.py => PodmanContainerDiff => 
```
        self.info = yaml.safe_load(json.dumps(info).lower())
        self.image_info = yaml.safe_load(json.dumps(image_info).lower())
```

my guess is that the reason is to have lower case config keys. But a bad side effect is that we also convert the values itself to lowercase. Now I have 2 scenarios where this will not work.

1. We have a config value like a volume path where Uppercase/Lowercase matters and we are not able to change it from uppercase to lowercase. In this case, we are not detecting any changes, because the original config is already converted to lowercase before we compare and the new value is lowercase too.

2. The Old value is Uppercase and the new value is uppercase, so no changes. In this case we detect always changes, because the old value is converted to lowercase before we compare it with the new value.

My question is now, what could be a proper fix. My recommendation is to convert only the keys to lowercase, but not the values.

sshnaidm:
> My question is now, what could be a proper fix. My recommendation is to convert only the keys to lowercase, but not the values.

No objections, the reason behind is different key cases in different podman versions, so lowering keys should be enough. Patches are welcome :)

HolgerHees:
ok. will create a tomorrow :-)

HolgerHees:
was not able to wait ;-)

sshnaidm:
Fixed by #115

PR #115, opened 2020-10-05
fix diff calculation

HolgerHees:
will fix this issues #113


sshnaidm:
Thanks, @HolgerHees, amazing! Let me test this with older versions of podman and centos and I'll merge it.

sshnaidm (changes requested):
In part of files (except of podman_container) `yaml` is not used then, so better to remove the import as well.

sshnaidm (approved):

Issue #94, opened 2020-07-27
Create podman logout module


sshnaidm:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Create podman logout module. which logs out from container registries.

Issue #93, opened 2020-07-27
Create podman login module


sshnaidm:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

Create podman module that logs in to container registries.

cfelder:
containers.podman.podman_image supports username and password.

isn't this sufficient?

sshnaidm:
Probably for most of cases, but I'd like to have it separate as well for one time login for the whole session.

istolpov:
@sshnaidm hi, Sergey.

No, that is not enough, since there is an ongoing community process of migration to podman.

Consider this:

1) Company has a role that performs initial setup (including docker login)
2) Company also has hundreds of roles for different containers, image pushes, whatever else.

Now, with your approach that's potentially hundreds upon hundreds lines of code in hundreds of files to add credentials to which is a) cumbersome, b) clutter.

With a login module that's 5 LOC of change in one file.

sshnaidm:
@istolpov I'm only for creating such module :smile:

sshnaidm:
Docs: https://containers.github.io/ansible-podman-collections/podman_login_module.html#ansible-collections-containers-podman-podman-login-module

PR #133, opened 2020-10-25
Add podman_login_info module

clelange:
This adds a module to check if one is logged in to a given container registry wrapping the `podman login --get-login` command with all its options.

I've also added a few tests, which are possible without an actual log in. The GitHub Actions workflow as well as the playbook are adjusted copies of the `podman_volume_info` equivalent.

I'd be willing to also look into creating `podman_login` and `podman_logout` modules as listed in #93 and #94 since they'd be useful for a project of mine.

Please let me know if you have any suggestions.

Labelling this PR for hacktoberfest would be appreciated.

### DIFF
--- a/plugins/modules/podman_login_info.py
+++ b/plugins/modules/podman_login_info.py
@@ -1,0 +1,117 @@
+#!/usr/bin/python
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+
+DOCUMENTATION = r"""
+module: podman_login_info
+author:
+  - "Clemens Lange (@clelange)"
+version_added: '1.0.0'
+short_description: Return the logged-in user if any for a given registry
+notes: []
+description:
+  - Return the logged-in user if any for a given registry.
+requirements:
+  - "Podman installed on host"
+options:
+  registry:
+    description:
+      - Registry server.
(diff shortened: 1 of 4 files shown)


plugins/modules/podman_login_info.py
```diff
@@ -0,0 +1,117 @@
...
+    result["logged_in"] = True
+    return result
+
+
+def main():
+    module = AnsibleModule(
+        argument_spec=dict(
+            executable=dict(type='str', default='podman'),
+            authfile=dict(type='path'),
+            registry=dict(type='str')
```
sshnaidm, line 97: I'd add registry as required argument because podman's "--get-login" expects for it. Although it doesn't fail when it's not specified, but also doesn't print registry, only username. I'm not sure only username without registry would be useful output for user.

sshnaidm (commented):
Thanks for pull request! One nit inside, please let me know what you think.

plugins/modules/podman_login_info.py
clelange: OK, makes sense! I'll look into it.

sshnaidm: cool, also please rebase on master when you submit, thanks

clelange:
Rebased and squashed into a single commit. Now the registry parameter is mandatory and I also added a test for it.

sshnaidm (approved):

sshnaidm:
@clelange very cool, thanks!

PR #134, opened 2020-10-25
Add podman logout module

clelange:
This adds a module to log out of a specified container registry or all registries wrapping the `podman logout` command with all its options.

I've also added a few tests, which are possible without an actual log in. The GitHub Actions workflow as well as the playbook are adjusted copies of the podman_volume_info equivalent.

Please let me know if you have any suggestions. Once login and login_info modules are in place, a combination of them can be tested together (if secrets are added to this repository).

Labelling this PR for hacktoberfest would be appreciated.

Addresses #94

### DIFF
--- a/.github/workflows/podman_logout.yml
+++ b/.github/workflows/podman_logout.yml
@@ -1,0 +1,110 @@
+name: Podman logout
+
+on:
+  push:
+    paths:
+      - '.github/workflows/podman_logout.yml'
+      - 'ci/*.yml'
+      - 'ci/run_containers_tests.sh'
+      - 'ci/playbooks/containers/podman_logout.yml'
+      - 'plugins/modules/podman_logout.py'
+      - 'tests/integration/targets/podman_logout/**'
+    branches:
+      - master
+  pull_request:
+    paths:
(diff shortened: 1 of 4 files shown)


sshnaidm:
Great, thanks!
While testing I discovered a case when you're logged in with docker tool and podman gives such output:
```
STDOUT:

Not logged into docker.io with current tool. Existing credentials were established via docker login. Please use docker logout instead.
```
Not sure what to in this case though, I think it should be `changed=false` because we didn't log out actually. But I don't think need to fail this module, maybe just to print a message somewhere.
Anyway, we can merge this as is and think about this special case later, I'm open to ideas.

clelange:
That's an interesting scenario. I didn't think of it.
I'd be in favour of failing the module in this case, since I'd expect that podman will still be able to use the credentials, therefore the sole purpose of this module, which is logging out, failed. Alternatively, one could add a flag to handle this case, e.g. `ignore_docker_credentials`, which would then prevent failure. What do you think?

sshnaidm:
@clelange yes, I think it's the best way to deal with it.

PR #137, opened 2020-10-27
Handle podman logout not logging out when logged in via different tool

clelange:
This is following up on the discussion in #134 

I've explained in the description in two places what is happening (here describing the new `ignore_docker_credentials` flag):

Credentials created using other tools such as `docker login` are not 
removed unless the corresponding `authfile` is explicitly specified.
Since podman also uses existing credentials in these files by default
(for docker e.g. `${HOME}/.docker/config.json`), module execution will
fail if a docker login exists for the registry specified in any
`authfile` is used by podman. This can be ignored by setting
`ignore_docker_credentials` to `yes`: the credentials will be kept and
`changed` will be false.
This option cannot be used together with `all` since in this case
podman will not check for existing `authfiles` created by other tools.

I've tested this with and without existing credentials using `docker login` and it seems to work as described.

### DIFF
--- a/plugins/modules/podman_logout.py
+++ b/plugins/modules/podman_logout.py
@@ -19,6 +19,10 @@ description:
     `authfile` flag. The default path used is
     `${XDG_RUNTIME_DIR}/containers/auth.json`.
     All the cached credentials can be removed by setting the `all` flag.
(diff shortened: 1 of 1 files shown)


sshnaidm (approved):
Looks great, thanks!

PR #181, opened 2021-01-20
Fix 'cap_add' and 'cap_drop' idempotency

gvalkov:
This change fixes containers always being rebuilt when the `cap_add` or `cap_drop` options are used. For example:

```yaml
  cap_add:
    - "NET_BIND_SERVICE"
    - "CAP_AUDIT_WRITE"
  ...
```

Consecutive runs produce the following diff:

```
--- before
+++ after
@@ -1 +1 @@
-cap_add - ['CAP_AUDIT_WRITE', 'CAP_NET_BIND_SERVICE']
+cap_add - ['CAP_AUDIT_WRITE', 'CAP_NET_BIND_SERVICE', 'cap_cap_audit_write', 'cap_net_bind_service']
```

Lower-casing and checking if the option already contains the `cap_` prefix fixes the issue (for me).

### DIFF
--- a/plugins/module_utils/podman/podman_container_lib.py
+++ b/plugins/module_utils/podman/podman_container_lib.py
@@ -670,21 +670,27 @@ class PodmanContainerDiff:
 
     def diffparam_cap_add(self):
         before = self.info['effectivecaps'] or []
+        before = [i.lower() for i in before]
         after = []
         if self.module_params['cap_add'] is not None:
-            after += ["cap_" + i.lower()
-                      for i in self.module_params['cap_add']]
+            for cap in self.module_params['cap_add']:
+                cap = cap.lower()
+                cap = cap if cap.startswith('cap_') else 'cap_' + cap
+                after.append(cap)
         after += before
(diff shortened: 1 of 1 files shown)


plugins/module_utils/podman/podman_container_lib.py
```diff
@@ -670,21 +670,27 @@ def diffparam_blkio_weight_device(self):
...
     def diffparam_cap_add(self):
         before = self.info['effectivecaps'] or []
+        before = [i.lower() for i in before]
         after = []
         if self.module_params['cap_add'] is not None:
-            after += ["cap_" + i.lower()
-                      for i in self.module_params['cap_add']]
+            for cap in self.module_params['cap_add']:
+                cap = cap.lower()
+                cap = cap if cap.startswith('cap_') else 'cap_' + cap
```
sshnaidm, line 678: hmm, I didn't think people are adding caps with starting CAP_, good point.

sshnaidm (approved):
LGTM, but it would be great to add tests about it.

Issue #182, opened 2021-01-29
log_level Not Idempotent on CentOS 8


derekmwright:
Formal bug report below, I have tested a fix and will attach PR shortly

**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**

1. On CentOS 8 install Podman 3.0.0-dev (current as of this issue creation)

2. Create a container using podman_container (see playbook below)

**Describe the results you received:**

Successive re-runs will always show a diff of:
```
--- before
+++ after
@@ -1 +1 @@
-log_level - warning
+log_level - warn
```

**Describe the results you expected:**

No change/diff

**Additional information you deem important (e.g. issue happens only occasionally):**

        "ExitCommand": [
            "/usr/bin/podman",
            "--root",
            "/var/lib/containers/storage",
            "--runroot",
            "/run/containers/storage",
            "--log-level",        <-- !!
            "warning",             <-- !!
            "--cgroup-manager",
            "systemd",
            "--tmpdir",
            "/run/libpod",
            "--runtime",
            "runc",
            "--storage-driver",
            "overlay",
            "--storage-opt",
            "overlay.mountopt=nodev,metacopy=on",
            "--events-backend",
            "file",
            "container",
            "cleanup",
            "e18babe156345f68b8e12afa0bac488be456e878b4a2d6e449b608804a23c9b7"
        ],
        ...
        "Config": {
            ...
            "CreateCommand": [
                "podman",
                "container",
                "run",
                "--name",
                "etcd",
                "--log-level",   <-- !!
                "warn",             <-- !!
                "--env",
                "ETCD_NAME=controlnode",
                "--env",
                "ETCD_LISTEN_CLIENT_URLS=http://0.0.0.0:2379",
                "--env",
                "ETCD_ADVERTISE_CLIENT_URLS=http://192.168.213.128:2379",
                "--env",
                "ETCD_LISTEN_PEER_URLS=http://0.0.0.0:2380",
                "--env",
                "ETCD_INITIAL_ADVERTISE_PEER_URLS=http://192.168.213.128:2380",
                "--publish",
                "2379:2379",
                "--publish",
                "2380:2380",
                "--detach=True",
                "quay.io/coreos/etcd:latest"
            ],
            ...

**Output of `ansible --version`:**

```
ansible 2.10.5
  config file = None
[...]

PR #183, opened 2021-01-29
Resolves Issue #182 - log_level not idempotent on CentOS 8

derekmwright:
This PR fixes the location the module looks in to determine the containers log-level setting when it was created. Prior to this fix it was checking the `exitcommand`, this is post container exit clean up commands for podman. Changed the function to look instead at `['config']['createcommand']` for the `log_level` setting to diff against. Ansible runs now execute without recreating the container every run.

### DIFF
--- a/plugins/module_utils/podman/podman_container_lib.py
+++ b/plugins/module_utils/podman/podman_container_lib.py
@@ -576,7 +576,7 @@ class PodmanDefaults:
             "ipc": "",
             "kernelmemory": "0",
             "log_driver": "k8s-file",
-            "log_level": "error",
+            "log_level": None,
             "memory": "0",
             "memory_swap": "0",
             "memory_reservation": "0",
@@ -886,7 +886,7 @@ class PodmanContainerDiff:
         if '--log-level' in createcom:
             before = createcom[createcom.index('--log-level') + 1].lower()
         else:
-            before = self.params['log_level']
+            before = None
         after = self.params['log_level']
         return self._diff_update_and_compare('log_level', before, after)
 
--- a/plugins/module_utils/podman/podman_container_lib.py
+++ b/plugins/module_utils/podman/podman_container_lib.py
@@ -882,9 +882,9 @@ class PodmanContainerDiff:
         return self._diff_update_and_compare('log_driver', before, after)
 
     def diffparam_log_level(self):
-        excom = self.info['exitcommand']
-        if '--log-level' in excom:
-            before = excom[excom.index('--log-level') + 1].lower()
+        createcom = self.info['config']['createcommand']
+        if '--log-level' in createcom:
+            before = createcom[createcom.index('--log-level') + 1].lower()
         else:
             before = self.params['log_level']
         after = self.params['log_level']


derekmwright:
Didn't test when log_level isn't set 😬 - working on it now.

derekmwright:
This one got a little interesting on how to handle a missing `--log-level` configuration flag. While it is true that Podman will default the value to `error` this is not reflected in the output for us to diff against. Therefore, the default value of a missing `--log-level` flag when checking the config should be `None`. This allows us to properly leave the a missing `log_level` ansible parameter alone and not set anything extra in the Podman container config and also properly detects a `log_level` config setting getting changed and will update to reflect the change. Not sure how to handle that on the documentation side or to just leave it alone.

derekmwright:
Going to look for guidance, further discussion on resolving the failing tests:

```
TASK [podman_container_idempotency : Run container with default log level] *****
task path: /home/runner/work/ansible-podman-collections/ansible-podman-collections/tests/integration/targets/podman_container_idempotency/tasks/idem_all.yml:110
<localhost> ESTABLISH LOCAL CONNECTION FOR USER: runner
<localhost> EXEC /bin/sh -c 'echo ~runner && sleep 0'
<localhost> EXEC /bin/sh -c '( umask 77 && mkdir -p "` echo /home/runner/.ansible/tmp `"&& mkdir "` echo /home/runner/.ansible/tmp/ansible-tmp-1611940745.1846876-47665-24629420274222 `" && echo ansible-tmp-1611940745.1846876-47665-24629420274222="` echo /home/runner/.ansible/tmp/ansible-tmp-1611940745.1846876-47665-24629420274222 `" ) && sleep 0'
Using module file /home/runner/.ansible/collections/ansible_collections/containers/podman/plugins/modules/podman_container.py
<localhost> PUT /home/runner/.ansible/tmp/ansible-local-28099gt1vz2az/tmp23uupoqm TO /home/runner/.ansible/tmp/ansible-tmp-1611940745.1846876-47665-24629420274222/AnsiballZ_podman_container.py
<localhost> EXEC /bin/sh -c 'chmod u+x /home/runner/.ansible/tmp/ansible-tmp-1611940745.1846876-47665-24629420274222/ /home/runner/.ansible/tmp/ansible-tmp-1611940745.1846876-47665-24629420274222/AnsiballZ_podman_container.py && sleep 0'
<localhost> EXEC /bin/sh -c '/opt/hostedtoolcache/Python/3.7.9/x64/bin/python /home/runner/.ansible/tmp/ansible-tmp-1611940745.1846876-47665-24629420274222/AnsiballZ_podman_container.py && sleep 0'
<localhost> EXEC /bin/sh -c 'rm -f -r /home/runner/.ansible/tmp/ansible-tmp-1611940745.1846876-47665-24629420274222/ > /dev/null 2>&1 && sleep 0'
--- before
+++ after
@@ -1 +1 @@
-log_level - error
+log_level - None
```

IMO this test should show changed as the removal of the log_level param should wipe out the explicit --log-level flag from the podman invocation. But I want to get input from regular contributors before I run around and start mucking with tests.

sshnaidm:
Sorry for delay, I'm fixing this in #186. We have a mechanism to overwrite defaults depending on current Podman version. Because every major (and not only) version something changes. The solution is just to add a required parameter to `default_dict` function here:
https://github.com/containers/ansible-podman-collections/blob/b911939f074c069d8625ad40dcb468fa19b71a15/plugins/module_utils/podman/podman_container_lib.py#L598

Like:
https://github.com/containers/ansible-podman-collections/blob/b911939f074c069d8625ad40dcb468fa19b71a15/plugins/module_utils/podman/podman_container_lib.py#L604-L605

Thanks for your effort, sorry it wasn't clear from code, it requires more documentation.
Hopefully we can proceed with different Podman versions in CI tests in the future.

derekmwright:
@sshnaidm thanks for the update! However I think there is still the outstanding issue that you are checking the configuration of the pod in the ExitCommand instead of the CreateCommand. ie: `['config']['createcommand']`

sshnaidm:
@derekmwright unfortunately "create command" is available only from v2. We need still to support 1.6.4 for some time.
But `log_level` seems to exist in exit command, what is the issue?

derekmwright:
> @derekmwright unfortunately "create command" is available only from v2. We need still to support 1.6.4 for some time.
> But `log_level` seems to exist in exit command, what is the issue?

ExitCommand is actually just cleanup commands for the container and does not reflect flags given to podman when the container is created.

sshnaidm:
> ExitCommand is actually just cleanup commands for the container and does not reflect flags given to podman when the container is created.

That's not quite correct, `--log-level` in `ExitCommand` does reflect flag of created container.

Issue #202, opened 2021-02-18
Allow containers to stop gracefully when recreating


eliasnorrby:
/kind feature

**Description**

Consider the following use case: we want to continually deploy an application (`app`) using the `latest` tag. Thus, we have the following task in our playbook:

```yaml
- name: Start container
  podman_container:
    name: "app"
    image: "app-image:latest"
    recreate: true
```

With `recrate: true`, the following actions will be performed:

```bash
podman rm -f app
podman run --name app app-image:latest
```

**Request**

Instead of forcefully removing the container, allow it to stop gracefully by first running `podman stop`.

sshnaidm:
fixed in #203

PR #203, opened 2021-02-18
Attempt graceful stop when recreating container

eliasnorrby:
See #202


sshnaidm:
@eliasnorrby thanks, LGTM, just let me fix CI after the Podman v3 released.

PR #236, opened 2021-04-05
Podman login module

jthiatt:
This is my first attempt at an Ansible module.  I have a workflow that could use a podman login task, so I thought I would give this a try.  Let me know if there is anything I missed or need to improve upon.

### DIFF
--- a/plugins/modules/podman_login.py
+++ b/plugins/modules/podman_login.py
@@ -102,7 +102,7 @@ def login(module, executable, registry, authfile,
         command.append(registry)
     if certdir:
         command.extend(['--cert-dir', certdir])
-    if tlsverify:
+    if tlsverify is not None:
         command.extend(['--tls-verify', tlsverify])
     rc, out, err = module.run_command(command)
     if rc != 0:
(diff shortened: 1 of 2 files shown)


tests/integration/targets/podman_login/tasks/main.yml
```diff
@@ -0,0 +1,33 @@
+- name: Test podman_login
```
sshnaidm, line 1: This will require additional configs in CI, but I'll handle it in next pr.

plugins/modules/podman_login.py
```diff
@@ -0,0 +1,160 @@
...
+        command.extend(['--username', username])
+    if password:
+        command.extend(['--password', password])
+    if authfile:
+        command.extend(['--authfile', authfile])
+    if registry:
+        command.append(registry)
+    if certdir:
+        command.extend(['--cert-dir', certdir])
+    if tlsverify:
```
sshnaidm, line 105: With booleans we should be careful and check if they are not `None`, otherwise `tlsverify=False` will never be configured.
```suggestion
    if tlsverify is not None:
```

sshnaidm (changes requested):
Thanks, looks great, just one comment

sshnaidm:
Please add "Fix #93" to PR description as it closes the issue.

plugins/modules/podman_login.py
jthiatt: Good catch, thank you

sshnaidm (approved):

sshnaidm:
Thanks!

PR #258, opened 2021-05-14
Fix multi-containers options

sshnaidm:
Options in multiple podman containers module are broken when they're lists or booleans. Ansible creates a list from a single option if it's defined as a list in args. Do the same, and convert option to the required type if need.

### DIFF
--- a/plugins/module_utils/podman/podman_container_lib.py
+++ b/plugins/module_utils/podman/podman_container_lib.py
@@ -128,6 +128,50 @@ ARGUMENTS_SPEC_CONTAINER = dict(
 )
 
 
+def init_options():
+    default = {}
+    opts = ARGUMENTS_SPEC_CONTAINER
+    for k, v in opts.items():
+        if 'default' in v:
+            default[k] = v['default']
+        else:
+            default[k] = None
+    return default
+
+
+def update_options(opts_dict, container):
+    def to_bool(x):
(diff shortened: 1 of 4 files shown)


sshnaidm:
@mwhahaha  please take a look, testing here: https://review.rdoproject.org/r/c/testproject/+/33683

sshnaidm:
github-check

sshnaidm:
github-check

sshnaidm:
check-github

mwhahaha (changes requested):
Looks ok, i've been testing something similar pulled into my module directly.
 https://review.opendev.org/c/openstack/tripleo-ansible/+/791317/7/tripleo_ansible/ansible_plugins/modules/tripleo_container_manage.py@153

Not certain if we want to include this here or make the expectation that these things would be handled by something outside of this.

sshnaidm:
> Looks ok, i've been testing something similar pulled into my module directly.
> https://review.opendev.org/c/openstack/tripleo-ansible/+/791317/7/tripleo_ansible/ansible_plugins/modules/tripleo_container_manage.py@153
> 
> Not certain if we want to include this here or make the expectation that these things would be handled by something outside of this.

Well, arguments types issue is definitely a bug in `podman_containers` module and should be fixed here. I think functions `_container_opts_defaults`, `_container_opts_update` and `_container_opts_types` can be handled here.

PR #262, opened 2021-05-24
Add support for timezones in containers

sshnaidm:


plugins/modules/podman_container.py
```diff
@@ -125,9 +125,6 @@
         The disabled option will force the container to not create CGroups,
```
inntran, line 125: looks like we need to update the description as well, or should we include a link to podman-run manpage on the top of the doc page?

PR #279, opened 2021-08-02
Fix idempotency issues

sshnaidm:
Fix idempotency when containers has a common network


ssbarnea:
I guess bit more work is needed...

Issue #291, opened 2021-08-29
podman_pod infra_name parameter


mskarbek:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

podman 3.3.0 added `--infra-name` to the `pod create` command, allowing custom name for infra container:
- release notes: https://github.com/containers/podman/releases/tag/v3.3.0
- pull request: https://github.com/containers/podman/pull/10820

PR #292, opened 2021-08-30
Add support for podman pod create --infra-name

mskarbek:
Fixes: #291

### DIFF
--- a/plugins/modules/podman_pod.py
+++ b/plugins/modules/podman_pod.py
@@ -100,6 +100,11 @@ options:
     - The image that will be created for the infra container. Default is "k8s.gcr.io/pause:3.1".
     type: str
     required: false
+  infra_name:
+    description:
+    - The name that will be used for the pod's infra container.
+    type: str
+    required: false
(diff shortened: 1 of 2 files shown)


plugins/modules/podman_pod.py
```diff
@@ -100,6 +100,11 @@
     - The image that will be created for the infra container. Default is "k8s.gcr.io/pause:3.1".
     type: str
     required: false
+  infra_name:
+    description:
+    - The name that will be used for the pod’s infra container.
```
sshnaidm, line 105: `pod’s` throws an error:

```diff
ERROR: plugins/modules/podman_pod.py:105:45: use ASCII quotes `'` and `"` instead of Unicode quotes
```

sshnaidm (changes requested):

plugins/modules/podman_pod.py
mskarbek: Fixed

sshnaidm (approved):

PR #299, opened 2021-09-14
Add systemd generation for pods

### LINKED ISSUE #18: create systemd unit files
Hi,
I am interested in implementing systemd units for containers and especially pods, would you see that as part of podman_container or should that go into a new module?

I already looked into the podman command, I think `podman generate systemd` needs another output format for it to be easily parseable in python, but I would welcome any suggestions.

Greetings
Klaas

sshnaidm:
Complete fix #18


StopMotionCuber:
I'd like to do a short follow up to this PR:
I just tried to replace my roles with this (from manual generation via `command`), and I noticed that there is no `.service` file  extension suffixed. I wanted to ask whether that was a deliberate design decision or rather something that was not thought of? 
With that extension, systemd would properly recognize service files placed into `/etc/systemd/system` and allow for an easier integration into a role

sshnaidm:
@StopMotionCuber will be solved via #309

PR #305, opened 2021-09-18
Fix suboption key in podman_container -> generate_systemd documentation

Bullja:
As of this line https://github.com/containers/ansible-podman-collections/blob/5b3f8a4d3dee5031bd639965539158ce273e9f84/plugins/module_utils/podman/common.py#L66 it should be "path", not "file"

### DIFF
--- a/plugins/modules/podman_container.py
+++ b/plugins/modules/podman_container.py
@@ -297,7 +297,7 @@ options:
     type: dict
     default: {}
     suboptions:
-      file:
+      path:
         description:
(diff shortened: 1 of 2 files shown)


sshnaidm:
Thanks, can you please fix in pod as well in the same patch?
https://github.com/containers/ansible-podman-collections/blob/5b3f8a4d3dee5031bd639965539158ce273e9f84/plugins/modules/podman_pod.py#L79-L85

Issue #285, opened 2021-08-09
Podman image push only, seems to not do so


sean-m-sullivan:
/kind bug

**Description**
Using Podman images to push an built ansible ee image, and it tries to pull image an image, even though that is set to false. 

**Steps to reproduce the issue:**

1.
```
  containers.podman.podman_image:
    auth_file:
    build:
      annotation:
      cache: true
      extra_args:
      force_rm: false
      format: oci
      rm: true
      volume:
    ca_cert_dir:
    executable: podman
    force: false
    name: ee_tools
    password: VALUE_SPECIFIED_IN_NO_LOG_PARAMETER
    path:
    pull: false
    push: true
    push_args:
      compress:
      dest: quay.io/ssulliva/ee_tools
      format:
      remove_signatures:
      sign_by:
      transport:
    state: present
    tag: latest
    username: ssulliva
    validate_certs:
 ``
I get the message
msg: Failed to pull image ee_tools:latest

The local podman images
```
podman images
REPOSITORY                            TAG      IMAGE ID      CREATED         SIZE
quay.io/ssulliva/ee_tools             latest   1a1e85769616  4 hours ago     737 MB
localhost/ee_tools                    latest   1a1e85769616  4 hours ago     737 MB
```
**Describe the results you expected:**
Expect it to be pushed, but definitely do not expect to get an error about pulling when trying to push a prebuilt image.


**Version of the `containers.podman` collection:**
```
containers.podman          1.6.2
```

**Output of `ansible --version`:**

```
ansible [core 2.11.2]
```

**Output of `podman version`:**

```
(paste your output here)
```
podman version 3.0.1
```
(paste your output here)
```

**Package info (e.g. output of `rpm -q podman` or `apt list podman`):**

```
(paste your output here)
```

**Playbok you run with ansible (e.g. content of `playbook.yaml`):**

```
(paste your output here)
```

**Comment by sshnaidm** [comment]
@sean-m-sullivan please provide exact playbook/task you are running for reproducing, thanks.

**Comment by sean-m-sullivan** [comment]
This is the playbook
```
---
- name: Playbook to configure ansible controller organizations
  hosts: localhost
  gather_facts: false
  collections:
    - sean_m_sullivan.tower_utilities
    - containers.podman
  vars:
    builder_dir: /tmp
    ee_registry_dest: quay.io/ssulliva/ee_tools
    ee_name: ee_tools 
    ee_continaer_runtime: docker
    ee_python:
      - podman
[...]

PR #327, opened 2021-10-17
Update error message when pull set to false

sean-m-sullivan:
In regards to #285 
I went through a few iterations of where to put this, while having minimal impact to functionality. This solution I think is the best option. The functionality of searching for the local image is preserved, however when pull is set to false. 

Instead of the current error, 
```
"msg": "Failed to pull image image_name:latest"
```
it will error that it failed to find the image locally. 
```
"msg": "Failed to find image image_name:latest locally, image pull set to False"
```
So that the user understands that the module failed to find the image. When pull is not set, or set to true, The original error message will display. 

I am open to any tweaks, but after banging my head against this error trying to find out what was going on, Just want to prevent any future user misunderstanding the error.

### DIFF
--- a/plugins/modules/podman_image.py
+++ b/plugins/modules/podman_image.py
@@ -542,7 +542,11 @@ class PodmanImageManager(object):
 
         rc, out, err = self._run(args, ignore_errors=True)
         if rc != 0:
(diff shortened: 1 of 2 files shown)


sean-m-sullivan:
Looking at the tests, Think I have one, will add and fix linting.

Issue #330, opened 2021-11-01
podman_secret adds \n to secret


wabuMike:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

The plugin containers.podman.podman_secret adds a newline (\n) to secrets provided with the parameter `data`.

**Steps to reproduce the issue:**

1. Add a task creating a secret with `podman secret`:

```yml
- name: Create secret with shell
  shell: "printf test123 | podman secret create TEST_1 -"
```

2. Add another task creating a secret with `containers.podman.podman_secret`:

```yml
- name: Create secret with podman_secret
  containers.podman.podman_secret:
    state: present
    name: TEST_2
    data: test123
```

3. Run a container with the secrets:

`podman run --rm -it --name alpine-test --secret TEST_1 --secret TEST_2 alpine`

4. Print  secrets:

```sh
/ # cat -e /run/secrets/TEST_1
test123/ # 
```

```sh
/ # cat -e /run/secrets/TEST_2
test123$
/ # 
```

**Describe the results you expected:**

When creating a secret with containers.podman.podman_secret, I expect the secret to have the exact same characters as I define with the parameter 'data'. In the example above, I would expect `cat -e /run/secrets/TEST_2` to print `test123/ #` as it does for `cat -e /run/secrets/TEST_1`.

**Additional information you deem important (e.g. issue happens only occasionally):**

**Version of the `containers.podman` collection:** 1.8.1
**Either git commit if installed from git: `git show --summary`**
**Or version from `ansible-galaxy` if installed from galaxy: `ansible-galaxy collection list | grep containers.podman`**

```
git show --summary
commit a0377fbd240f39d35c8b833b1f961c61b445fc1b (HEAD -> master, origin/master, origin/HEAD)
Author: Oliver Walsh <owalsh@redhat.com>
Date:   Tue Oct 19 15:18:22 2021 +0100
```

**Output of `ansible --version`:**

```
ansible 2.10.2
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/michael/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/michael/.local/lib/python3.8/site-packages/ansible
  executable location = /home/michael/.local/bin/ansible
  python version = 3.8.10 (default, Sep 28 2021, 16:10:42) [GCC 9.3.0]
```

**Output of `podman version`:**

```
Version:      3.3.1
[...]

PR #331, opened 2021-11-01
Don't add newlines to secrets

wabuMike:
This solves issue https://github.com/containers/ansible-podman-collections/issues/330 , which describes how podman_secret adds newlines (\n) to secrets.

After having a look at the [source code](https://github.com/containers/ansible-podman-collections/blob/a0377fbd240f39d35c8b833b1f961c61b445fc1b/plugins/modules/podman_secret.py#L99) and the [run_command documentation](https://docs.ansible.com/ansible/latest/reference_appendices/module_utils.html#ansible-module-utils), it seems the parameter binary_data of run_command must be `true` to prevent new lines being added to data:

> Kw binary_data - If False, append a newline to the data. Default False

### DIFF
--- a/plugins/modules/podman_secret.py
+++ b/plugins/modules/podman_secret.py
@@ -96,7 +96,7 @@ def podman_secret_create(module, executable, name, data, force, skip):
             }
 
     rc, out, err = module.run_command(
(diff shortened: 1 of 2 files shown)


sshnaidm:
Thanks! Nice catch and fix!

PR #332, opened 2021-11-09
Allow to actually pass a list of string for "mounts"

cjeanner:
Like "volume", "mount" can be passed multiple times to the container
creation/run, therefore we want to support a list of string, and a
"mounts" alias.

### DIFF
--- a/plugins/module_utils/podman/podman_container_lib.py
+++ b/plugins/module_utils/podman/podman_container_lib.py
@@ -91,7 +91,7 @@ ARGUMENTS_SPEC_CONTAINER = dict(
     memory_reservation=dict(type='str'),
     memory_swap=dict(type='str'),
     memory_swappiness=dict(type='int'),
-    mount=dict(type='str'),
+    mount=dict(type='list', elements='str', aliases=['mounts']),
     network=dict(type='list', elements='str', aliases=['net', 'network_mode']),
     network_aliases=dict(type='list', elements='str'),
     no_hosts=dict(type='bool'),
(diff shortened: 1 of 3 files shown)


sshnaidm:
Do you mind to add test for that? Like here: https://github.com/containers/ansible-podman-collections/blob/master/tests/integration/targets/podman_container/tasks/main.yml#L356-L357
Also need to add idempotency calculation, I'll try to handle this.

sshnaidm:
Also need to change it here: https://github.com/containers/ansible-podman-collections/blob/master/plugins/module_utils/podman/podman_container_lib.py#L94

sshnaidm (changes requested):
/ignore it

sshnaidm (approved):

cjeanner:
added a test for the "mounts" option. Let's see. Apparently, podman puts the "mount" things before "volume". I tested the "podman inspect" output from this command:
```
podman run --volume /dev:/dev --mount type=devpts,destination=/dev/pts --rm -ti centos:8 bash
```
We get:
```
        "Mounts": [
            {
                "Type": "bind",
                "Source": "devpts",
                "Destination": "/dev/pts",
                "Driver": "",
                "Mode": "",
                "Options": [],
                "RW": true,
                "Propagation": ""
            },
            {
                "Type": "bind",
                "Source": "/dev",
                "Destination": "/dev",
                "Driver": "",
                "Mode": "",
                "Options": [
                    "nosuid",
                    "rbind"
                ],
                "RW": true,
                "Propagation": "rprivate"
            }
        ],
```
And we get the same output when we switch the --volume and --mount options.

Issue #337, opened 2021-11-18
AttributeError: module 'signal' has no attribute 'SIGRTMAX'


pschiffe:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Latest (1.8.2) version doesn't work on macOS, example code:
```
- name: pull an image
  containers.podman.podman_image:
    name: postgres:{{ pgsql_version }}-alpine
    state: present
```
fails with `AttributeError: module 'signal' has no attribute 'SIGRTMAX'`

Target is VM with CentOS 8.5. Ansible is running on macOS 11 with Intel CPU.

Version 1.8.1 is working fine for me. Problem was probably introduced with PR #326 
cc @olliewalsh 

**Version of the `containers.podman` collection:**
**Either git commit if installed from git: `git show --summary`**
**Or version from `ansible-galaxy` if installed from galaxy: `ansible-galaxy collection list | grep containers.podman`**

```
1.8.2
```

**Output of `ansible --version`:**

```
ansible 2.9.27
  config file = /Users/.../ansible.cfg
  configured module search path = ['/Users/pschff/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/local/Cellar/ansible@2.9/2.9.27_1/libexec/lib/python3.9/site-packages/ansible
  executable location = /usr/local/opt/ansible@2.9/bin/ansible
  python version = 3.9.8 (main, Nov 10 2021, 09:21:22) [Clang 13.0.0 (clang-1300.0.29.3)]
```

**Output of `podman version`:**

```
Version:      3.3.1
API Version:  3.3.1
Go Version:   go1.16.7
Built:        Tue Nov  9 22:23:56 2021
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
  buildahVersion: 1.22.3
  cgroupControllers:
  - cpuset
  - cpu
  - cpuacct
  - blkio
  - memory
  - devices
  - freezer
  - net_cls
  - perf_event
  - net_prio
  - hugetlb
  - pids
  - rdma
  cgroupManager: systemd
  cgroupVersion: v1
  conmon:
    package: conmon-2.0.29-1.module_el8.5.0+890+6b136101.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.0.29, commit: 84384406047fae626269133e1951c4b92eed7603'
  cpus: 2
[...]

PR #338, opened 2021-11-19
Hardcode RT signal numbers

olliewalsh:
Looking up SIGRTMIN/SIGRTMAX isn't portable.
Hardcode the Linux signal number instead.

Fixes issue #337.


olliewalsh:
FWIW [man 7 signal](https://man7.org/linux/man-pages/man7/signal.7.html) states "programs should never refer to real-time signals using hard-coded numbers". However podman is hardcoding it.

PR #344, opened 2021-11-26
Add secrets driver and driver opts support

amenzhinsky:


tests/integration/targets/podman_secret/tasks/main.yml
```diff
@@ -73,6 +73,20 @@
         that:
           - removed is not changed
 
+    - name: Create secret with file driver and custom options
```
amenzhinsky, line 76: Unfortunately we cannot check correctness of the test, but at least it shows that podman doesn't fail. 

I checked it manually with `cat ~/.local/share/containers/storage/secrets/secrets.json` to verify that driver and options are the same that we requested.

sshnaidm:
LGTM

PR #346, opened 2021-11-27
Add a second example to podman_pod_module.html

Tronde:
Hi, What do you think about adding a second example to the example section of podman_pod_module.html?


sshnaidm (changes requested):
Documentation is generated automatically from YAML files. Please add it here: https://github.com/containers/ansible-podman-collections/blob/master/plugins/modules/podman_pod.py#L282-L288
And next time I'll re-generate the docs.
Either you can play with this script to generate them by yourself: https://github.com/containers/ansible-podman-collections/blob/master/contrib/build_docs.sh

Tronde:
Hi @sshnaidm,  
Thanks for your reply. I did undo the changes in `docs/podman_pod_module.html` and add the example to `plugins/modules/podman_pod.py` as you suggested.

I did not try the script to generate the docs. I would appreciate if you do the honors.

sshnaidm:
LGTM

sshnaidm (approved):

PR #359, opened 2021-12-31
Fix podman_pod_lib behavior for ports published to multiple IPs

ttomasini:
Fix podman_pod_lib behavior if the same port has been published to multiple network IPs. This lead to the restart of the pod all the time because the `before` array only contained the first IP (`j[0]['hostip']`).

#### configuration
```yaml
- name: Add new pod
  containers.podman.podman_pod:
    name: pod-foobar
    ports:
      - 10.126.0.1:1234:1234
      - 127.0.0.1:1234:1234
```
#### playbook execution
```diff
TASK [container : Add new pod] ************************************************************************************************************************
--- before
+++ after
@@ -1 +1 @@
-publish - ['10.126.0.1:1234:1234']
+publish - ['10.126.0.1:1234:1234', '127.0.0.1:1234:1234']

changed: [server.hostname]
```

Additional remark: I would recommend to get rid of the duplicate code if there is some time left in the future to avoid issues like this.


sshnaidm:
Please see failures in pod job, need to fix ipv6 case:
https://github.com/containers/ansible-podman-collections/blob/fcf15221478e2075993a789cbf4c8056d8e01f18/tests/integration/targets/podman_pod/tasks/root-pod.yml#L128-L139

```
TASK [podman_pod : Start pod with ports for idempotency and ipv6 again] ********
task path: /home/runner/work/ansible-podman-collections/ansible-podman-collections/tests/integration/targets/podman_pod/tasks/root-pod.yml:128
--- before
+++ after
@@ -1 +1 @@
-publish - ['127.0.0.1:13122:8871', '127.0.0.1:43423:8872', '127.0.0.1:7671:7676/udp', '4444:4444', '8888:19191/udp', '::1:8743:8745']
+publish - ['127.0.0.1:13122:8871', '127.0.0.1:43423:8872', '127.0.0.1:7671:7676/udp', '1:8743:8745', '4444:4444', '8888:19191/udp']

```
Also please add there a test case with multiple network IPs.
Thanks!

ttomasini:
@sshnaidm thanks for your remarks. All issues are hopefully fixed, now. Please have a look and start the pipeline to see if the tests are running properly.
Happy new year and thanks for your great work so far!

tests/integration/targets/podman_pod/tasks/root-pod.yml
```diff
@@ -115,9 +115,11 @@
           - "4444:4444/tcp"
           - "8888:19191/udp"
           - "127.0.0.1:7671:7676/udp"
+          - "10.126.0.1:7671:7676/udp"
```
sshnaidm, line 118: It fails with error `Error: error starting container 0fd706662950af98a7b24891f529aaa1dedf7947988773123e16f9037c2b1059: cannot listen on the UDP port: listen udp4 10.126.0.1:7671: bind: cannot assign requested address`
I think it's better to use localhost addresses, like:

```yaml
- "127.0.0.1:7671:7676/udp"
- "127.0.0.2:7671:7676/udp"
 ```

File: tests/integration/targets/podman_pod/tasks/root-pod.yml [hunk 2]
```diff
@@ -115,9 +115,11 @@
...
           - "127.0.0.1:13122:8871/tcp"
           - "127.0.0.1:43423:8872"
           - "[::1]:8743:8745"
+          - "[fd42::1]:8743:8745"
```
sshnaidm, line 122: probably same here with `::2`

sshnaidm (changes requested):
Thanks, just need to fix test.

ttomasini:
@sshnaidm thanks for the second review. I fixed the binding issue for IPv4 but I guess there is no solution for IPv6 since there only exists a single loopback address which is [::1]. [::2] will probably run into the same binding error. But the test should still be sufficient enough.

sshnaidm (approved):
Thanks for contribution!

Issue #364, opened 2022-01-06
podman_login handles tlsverify incorrectly


rdrazny:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When setting the `tlsverify`  option to `False` for a `containers.podman.podman_login` task, the task fails with a python traceback.

**Steps to reproduce the issue:**

1. Create an Ansible playbook with a `containers.podman.podman_login` containing `tlsverify: No`

2. Run the playbook

**Describe the results you received:**

The task fails with the following output

```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: TypeError: expected str, bytes or os.PathLike object, not bool
[...]

PR #365, opened 2022-01-06
Handle tlsverify correctly in podman_login

rdrazny:
#364

### DIFF
--- a/plugins/modules/podman_login.py
+++ b/plugins/modules/podman_login.py
@@ -103,7 +103,10 @@ def login(module, executable, registry, authfile,
     if certdir:
         command.extend(['--cert-dir', certdir])
     if tlsverify is not None:
-        command.extend(['--tls-verify', tlsverify])
+        if tlsverify:
+            command.append('--tls-verify')
+        else:
+            command.append('--tls-verify=False')
     rc, out, err = module.run_command(command)
     if rc != 0:
         if 'Error: Not logged into' not in err:


plugins/modules/podman_login.py
```diff
@@ -103,7 +103,10 @@ def login(module, executable, registry, authfile,
     if certdir:
         command.extend(['--cert-dir', certdir])
     if tlsverify is not None:
-        command.extend(['--tls-verify', tlsverify])
+        if tlsverify:
+            command.append('--tls-verify')
```
sshnaidm, line 107: If I read it correctly, you'll need `=true` here:
from `man podman-login`:
```
--tls-verify=true|false
       Require HTTPS and verify certificates when contacting registries (default: true). If explicitly set to true, then TLS verification will be used. If set to false, then TLS verification  will  not  be
       used. If not specified, TLS verification will be used unless the target registry is listed as an insecure registry in registries.conf.
```

rdrazny: Well, checking the `podman login --help`, it is somewhat ambiguous (or not very well documented).
```
$ podman login --help
Login to a container registry

Description:
  Login to a container registry on a specified server.

Usage:
  podman login [options] [REGISTRY]

Examples:
  podman login quay.io
  podman login --username ... --password ... quay.io
  podman login --authfile dir/auth.json quay.io

Options:
      --authfile string   path of the authentication file. Use REGISTRY_AUTH_FILE environment variable to override
      --cert-dir string   use certificates at the specified path to access the registry
      --get-login         Return the current login user for the registry
  -p, --password string   Password for registry
      --password-stdin    Take the password from stdin
      --tls-verify        Require HTTPS and verify certificates when contacting registries
  -u, --username string   Username for registry
  -v, --verbose           Write more detailed information to stdout
```
There is no mention of `True|False`, and passing just`--tls-verify` implies `--tls-verify=True`. The man page is more detailed, though (while not completely clear too) , and explicitly says that `True` is default
```
   --tls-verify=true|false
       Require HTTPS and verify certificates when contacting registries (default: true). If explicitly set to true, then TLS verification will be used. If set to false, then TLS verification will not be used. If not specified,
       TLS verification will be used unless the target registry is listed as an insecure registry in registries.conf.
```
I have tested with and without a record for an insecure registry in registries.conf, and `--tls-verify` without '=true' on the command line overrides `insecure=true` option in `registries.conf`, so `true` is actually there, even if not explicitly specified.

sshnaidm: I see, thanks for checking, vague moment indeed.

Issue #383, opened 2022-03-04
podman_container does not recreate the container if root image is changed from read-only to read-write


rubiksdot:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

**Steps to reproduce the issue:**

1. Create container read-only
2. Comment out read-only: true
3. Attempt to recreate

**Describe the results you received:**

Read-only container remained.

**Describe the results you expected:**

Read-only container replaced with one whose root is read-write.

**Additional information you deem important (e.g. issue happens only occasionally):**

Found that read_only is not a setting tested for change so implemented this patch which appears to do the right thing. Changed params['read_only'] to default to False as per docs so that the comparison works if commented out.

```
--- plugins/module_utils/podman/podman_container_lib.py.old     2022-03-02 15:22:27.688204879 +1100
+++ plugins/module_utils/podman/podman_container_lib.py 2022-03-03 18:04:50.224838188 +1100
@@ -104,7 +104,7 @@
     publish=dict(type='list', elements='str', aliases=[
         'ports', 'published', 'published_ports']),
     publish_all=dict(type='bool'),
-    read_only=dict(type='bool'),
+    read_only=dict(type='bool', default=False),
     read_only_tmpfs=dict(type='bool'),
     recreate=dict(type='bool', default=False),
     restart_policy=dict(type='str'),
@@ -1150,6 +1150,11 @@
         before, after = sorted(list(set(before))), sorted(list(set(after)))
         return self._diff_update_and_compare('publish', before, after)
 
+    def diffparam_read_only(self):
+        before = self.info['hostconfig']['readonlyrootfs']
+        after = self.params['read_only']
+        return self._diff_update_and_compare('read_only', before, after)
+
     def diffparam_rm(self):
         before = self.info['hostconfig']['autoremove']
         after = self.params['rm']
```

**Version of the `containers.podman` collection:**
**Either git commit if installed from git: `git show --summary`**
**Or version from `ansible-galaxy` if installed from galaxy: `ansible-galaxy collection list | grep containers.podman`**

```
containers.podman         1.4.1  
containers.podman             1.5.0  
containers.podman 1.9.1  
```

(1.9.1 is in play)

**Output of `ansible --version`:**

```
ansible 2.10.17
  config file = /data/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
[...]

PR #388, opened 2022-03-09
bug 383: fix check for read-only change of root image

rubiksdot:
The --read-only option is not checked for changes between runs of
ansible thus the container is not recreated if the user changes
the root image between read-write to read-only.

Default for the read_only variable set to False in line with
documentation and so that later checks are simplified.

This is a fix for bug #383.


sshnaidm:
Please sign your commit: `git commit -s`, it's must in all repos in this org.

Issue #394, opened 2022-03-16
Pod: Can't publish containerPort <60


Henry78:
/kind bug

**Description**

Trying to create a pod with containerPort <60 published fails with `Error: error parsing container port`

**Steps to reproduce the issue:**

1. Run playbook that creates a pod with a published port, where the containerPort is <60 

**Describe the results you received:**
I receive the error below. Note, that the port in the error (`got 133343`) depends on the containerPort that's mapped/published. 
- :20 > got 133340
- :21 > got 133341
- :22 > got 133342
- :23 > got 133343
- :58 > got 133378
- :59 > got 133379
- :60 > no error- 
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Can't create pod demo", "stderr": "Error: error parsing container port: port numbers must be between 1 and 65535 (inclusive), got 133343\n", "stderr_lines": ["Error: error parsing container port: port numbers must be between 1 and 65535 (inclusive), got 133343"], "stdout": "", "stdout_lines": []}
```

**Log output**
```
21:36:52 shannon ansible-containers.podman.podman_pod[279188]: Invoked with name=demo state=started debug=True publish=['133343'] recreate=False generate_systemd={} executable=podman add_host=None cgroup_parent=None cpus=None cpuset_cpus=None device=None device_read_bps=None dns=None dns_opt=None dns_search=None gidmap=None hostname=None infra=None infra_conmon_pidfile=None infra_command=None infra_image=None infra_name=None ip=None label=None label_file=None mac_address=None network=None network_alias=None no_hosts=None pid=None pod_id_file=None share=None subgidname=None subuidname=None uidmap=None userns=None volume=None
21:36:52 shannon ansible-containers.podman.podman_pod[279188]: PODMAN-POD-DEBUG: podman pod create --name demo --publish 133343
```

**Describe the results you expected:**
Pod gets created.

**Version of the `containers.podman` collection:**
```
containers.podman             1.9.1  
```

**Output of `ansible --version`:**

```
ansible [core 2.12.3]
  config file = /etc/ansible/ansible.cfg
  configured module search path = ['/home/henry/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/henry/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
[...]

PR #401, opened 2022-03-23
Podman Pod: Update examples to use quotes when publishing ports

Henry78:
to avoid them being intrepreted as scalars. #394


sshnaidm:
Please sign your commit: `git commit -s`  it's must in all repos in this org.

PR #402, opened 2022-03-28
Fix log_level module parameter

cloudnull:
The `log_level` module parameter is not guaranteed by Ansible and not
something we have a default for within the library. If the `log_level`
parameter is not defined the following error will be raised resulting
in deployment failure [0]. This change adjusts how the `log_level`
parameter is collected so that we can ensure the module isn't failing
unnecessarily.

[0] https://paste.opendev.org/show/b3dJmgeuZKna74RQo5vy/

Signed-off-by: Kevin Carter <kecarter@redhat.com>

PR #403, opened 2022-03-28
Fixed handling of log_level module parameter

JM1:
Parameter 'log_level' might not have been defined in module parameters.
Use self.module_params instead of self.module.params because what is
passed to PodmanManager as params might differ from module params, e.g.
in module podman_containers.

### DIFF
--- a/plugins/module_utils/podman/podman_container_lib.py
+++ b/plugins/module_utils/podman/podman_container_lib.py
@@ -976,7 +976,7 @@ class PodmanContainerDiff:
         if '--log-level' in excom:
             before = excom[excom.index('--log-level') + 1].lower()
         else:
-            if self.module.params['log_level'] is not None:
+            if 'log_level' in self.module_params:
                 before = ''
             else:
                 before = self.params['log_level']


plugins/module_utils/podman/podman_container_lib.py
```diff
@@ -976,7 +976,7 @@ def diffparam_log_level(self):
         if '--log-level' in excom:
             before = excom[excom.index('--log-level') + 1].lower()
         else:
-            if self.module.params['log_level'] is not None:
+            if 'log_level' in self.module_params:
```
cloudnull, line 979: we still need to check if the value is not none.

sshnaidm:
`self.module.params['log_level']` ill be always defined, since it's defined here: 
https://github.com/sshnaidm/ansible-podman-collections/blob/55a2efa0d07782fb61838dab8e6116e9fb19396c/plugins/module_utils/podman/podman_container_lib.py#L81-L83
And if it's not set in module, it will be `None`.
What is the problem you see? Did you hit any bug? If so, can you open the issue?

plugins/module_utils/podman/podman_container_lib.py
sshnaidm: see my comment in PR

sshnaidm (changes requested):

JM1:
Not a bug in Podman collection, see https://github.com/containers/ansible-podman-collections/pull/402#issuecomment-1081563552 .

PR #404, opened 2022-03-29
Remove idempotency for log level

sshnaidm:
Since Podman changes its defaults very often and it's impossible
to track and identify every time what is current log level, let's
remove idempotency for this parameter for now.


sshnaidm:
check-github

Issue #384, opened 2022-03-04
podman_container does not recreate the container if image changes


rubiksdot:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

I changed the CMD in Containerfile and had the image recreated. podman_container failed to detect this and, so, left the now stale container configured.

**Steps to reproduce the issue:**

1. Create image
2. Create container based on image
3. Change Containerfile and recreate image
4. Run step 2.

**Describe the results you received:**

Container created in step 2 remained.

**Describe the results you expected:**

Container created in step 2 to have been recreated (or updated - don't care) in step 4.

**Additional information you deem important (e.g. issue happens only occasionally):**

I fixed this according to the diff below. End result is that it does not recreate the container when no changes are made to the image and does recreate it when I changed the Containerfile.

```
--- plugins/module_utils/podman/podman_container_lib.py.old     2022-03-02 15:22:27.688204879 +1100
+++ plugins/module_utils/podman/podman_container_lib.py 2022-03-02 15:22:29.968213862 +1100
@@ -916,7 +916,7 @@
     def diffparam_image(self):
         before_id = self.info['image']
         after_id = self.image_info['id']
-        if before_id == after_id:
+        if before_id != after_id:
             return self._diff_update_and_compare('image', before_id, after_id)
         before = self.info['config']['image']
         after = self.params['image']
```

**Version of the `containers.podman` collection:**
**Either git commit if installed from git: `git show --summary`**
**Or version from `ansible-galaxy` if installed from galaxy: `ansible-galaxy collection list | grep containers.podman`**

```
containers.podman         1.4.1  
containers.podman             1.5.0  
containers.podman 1.9.1  
```

(1.9.1 is in play)

**Output of `ansible --version`:**

```
ansible 2.10.17
  config file = /data/ansible/ansible.cfg
  configured module search path = ['/root/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /root/.local/lib/python3.9/site-packages/ansible
  executable location = /root/.local/bin/ansible
  python version = 3.9.2 (default, Feb 28 2021, 17:03:44) [GCC 10.2.1 20210110]
```

**Output of `podman version`:**

```
Version:      3.0.1
API Version:  3.0.0
Go Version:   go1.15.9
Built:        Thu Jan  1 10:00:00 1970
OS/Arch:      linux/amd64
```

[...]

PR #410, opened 2022-04-04
fix image change detection on id change

rubiksdot:
If the image id changes than the image changed. As such we should
return signalling that change.

This is a fix for bug #384.


sshnaidm:
Thanks, I'll test it asap.

sshnaidm (changes requested):
waiting for reply https://github.com/containers/ansible-podman-collections/issues/384#issuecomment-1108365573

sshnaidm:
Closing for now, please feel free to resubmit in case it's still an issue.

Issue #414, opened 2022-04-15
Inconsistent podman_container and podman_pod syntax - network attribute


MrSuicideParrot:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**
The podman_container module has the attribute [network](https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-network) as a list of strings. The podman_pod has the same attribute [network](https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_pod_module.html#parameter-network) as a string. In the case of pods, if we want to specify multiple networks we have to concatenate them with a comma. This causes incoherence in the API, at the same time is not normal to have this kind of syntax (comma separated strings) in ansible.

**Describe the results you expected:**
I expected that both parameters would be a list of strings.

sshnaidm:
Sounds reasonable, would you like to send a patch?

MrSuicideParrot:
Yes, I can do it.

PR #416, opened 2022-04-21
Change network attribute from str to list in pods #414

MrSuicideParrot:
This pull request implements the suggestion mentioned in the issue #414

### DIFF
--- a/plugins/modules/podman_load.py
+++ b/plugins/modules/podman_load.py
@@ -155,7 +155,7 @@ def load(module, executable):
     if rc != 0:
         module.fail_json(msg="Image loading failed: %s" % (err))
     image_name_line = [i for i in out.splitlines() if 'Loaded image' in i][0]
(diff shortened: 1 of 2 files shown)


sshnaidm:
Please sign your commit: `git commit -s`, it's must in all repos in this org.

MrSuicideParrot:
I checked the failed tasks, and I wasn't able to understand why they failed. The network `10.90.90.0/24` is only used in one place and is created by a podman command, so I don't understand how my commit managed to break this.  Do you have any idea?

sshnaidm (changes requested):
In line 495 need to change:

```diff
- if self.params['network'] == 'host':
+ if 'host' in self.params['network']:
```

sshnaidm:
The real problem is here: https://paste.opendev.org/show/bqf4dxKPr2QcoGLgoBAS/

sshnaidm:
Sorry, seems like I screwed it up while updating your patch. Can you resubmit maybe?

PR #417, opened 2022-04-24
 Change network attribute from str to list in pods #414 #416 

### LINKED ISSUE #414: Inconsistent podman_container and podman_pod syntax - network attribute

sshnaidm:
Fix #414
Fix #416

### DIFF
--- a/plugins/module_utils/podman/podman_pod_lib.py
+++ b/plugins/module_utils/podman/podman_pod_lib.py
@@ -46,7 +46,7 @@ ARGUMENTS_SPEC_POD = dict(
     label_file=dict(type='str', required=False),
     mac_address=dict(type='str', required=False),
     name=dict(type='str', required=True),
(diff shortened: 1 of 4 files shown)


sshnaidm:
@MrSuicideParrot resurrected it here

PR #430, opened 2022-05-26
Fix idem-potency for podman_login

### LINKED ISSUE #426: Shouldn't containers.podman.podman_login be idempotent?
Using the default molecule scenario which include an idempotency test, which checks that every task of a role returns change=false on a second run, I noticed that containers.podman.podman_login was returning change=true, even the auth.json file was already existing with the same credentials.

There might be reasons why the actual behavior is expected, but as far as I'm concerned with my usage it'd seem normal to receive a change=false when running containers.podman.podman_login with credentials already stored in auth.json.

clelange:
Fixes #426 

This is an attempt to ensure idem-potency for the login module. I decided to calculate a hash of the auth JSON to figure out if there's been a change since the login command writes the login information to this file even if it's reusing existing credentials. An alternative could be to open the auth file and base64-encoding the provided login information to check for differences. However, I don't think that would simplify the code.

### DIFF
--- a/plugins/modules/podman_login.py
+++ b/plugins/modules/podman_login.py
@@ -8,6 +8,7 @@ DOCUMENTATION = r'''
 module: podman_login
 author:
   - "Jason Hiatt (@jthiatt)"
+  - "Clemens Lange (@clelange)"
 short_description: Login to a container registry using podman
 notes: []
 description:
@@ -83,6 +84,8 @@ EXAMPLES = r"""
 # noqa: F402
 
 import json  # noqa: F402
+import hashlib
+import os
 from ansible.module_utils.basic import AnsibleModule
 
 
@@ -98,6 +101,9 @@ def login(module, executable, registry, authfile,
         command.extend(['--password', password])
     if authfile:
         command.extend(['--authfile', authfile])
+        authfile = os.path.expandvars(authfile)
+    else:
+        authfile = os.getenv('XDG_RUNTIME_DIR', '') + '/containers/auth.json'
     if registry:
         command.append(registry)
     if certdir:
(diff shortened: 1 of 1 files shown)


plugins/modules/podman_login.py
```diff
@@ -107,6 +113,18 @@ def login(module, executable, registry, authfile,
...
+    # Use a checksum to check if the auth JSON has changed
+    checksum = None
+    docker_authfile = os.path.expandvars('$HOME/.docker/config.json')
+    if os.path.exists(authfile):
+        content = open(authfile, 'rb').read()
+        if bytes(registry, 'UTF-8') in content:
+            checksum = hashlib.md5(content).hexdigest()
+    # podman falls back to ~/.docker/config.json if the default authfile doesn't exist
+    elif os.path.exists(docker_authfile):
+        content = open(docker_authfile, 'rb').read()
```
sshnaidm, line 125: Instead of duplication, you can discover a file before and then to calculate the checksum, something like:

```python
check_file = authfile if os.path.exists(authfile) else docker_authfile
if os.path.exists(check_file):
        content = open(authfile, 'rb').read()
        if bytes(registry, 'UTF-8') in content:
            checksum = hashlib.md5(content).hexdigest()

```

plugins/modules/podman_login.py
```diff
@@ -116,6 +134,21 @@ def login(module, executable, registry, authfile,
         changed = True
         if 'Existing credentials are valid' in out:
             changed = False
+        # If we have managed to calculate a checksum before, check if it has changed
+        # due to the login
+        if checksum:
+            if os.path.exists(authfile):
```
sshnaidm, line 140: same here, if we found a required file before, let's just use it:

```python
if checksum and os.path.exists(check_file):
    content = open(authfile, 'rb').read()
  ....
```

sshnaidm (commented):
Thanks, a few comments about the code.
What I wonder is - what if we logged in but didn't change any of files - could it be? Like timed out login maybe or other reason.

plugins/modules/podman_login.py
clelange: Thanks for the suggestion, that's indeed much better (with a small change to open `check_file`)

plugins/modules/podman_login.py
clelange: We don't need to check for the existence of the file since without its existence, we couldn't have calculated the checksum. I've implemented the change.

clelange:
> What I wonder is - what if we logged in but didn't change any of files - could it be? Like timed out login maybe or other reason.

I'm not quite sure I understand. I think the auth file will only be changed when the login has been successful. And a login means either using the existing credentials in the file or the provided log in information and updating the auth file (that contains a dictionary of the registry names and their base64-encoded user+pass) after successful login.

sshnaidm (approved):
Thanks!

jthiatt:
Thanks @clelange , I just haven't had a chance to look at this.

PR #433, opened 2022-05-31
Support for only setting mtu or vlan

Turmio:
Current `opt` parameter implementation requires that both mtu and vlan is defined. After this PR default value is used it other parameter is not defined.

### DIFF
--- a/plugins/modules/podman_network.py
+++ b/plugins/modules/podman_network.py
@@ -386,7 +386,7 @@ class PodmanNetworkDiff:
             vlan_before = self.info['plugins'][0].get('vlan')
         except (IndexError, KeyError):
             vlan_before = None
-        vlan_after = self.params['opt'].get('vlan') if self.params['opt'] else None
+        vlan_after = self.params.get('opt', {}).get('vlan', None)
         if vlan_before or vlan_after:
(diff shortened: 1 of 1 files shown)


plugins/modules/podman_network.py
```diff
@@ -386,7 +386,7 @@ def diffparam_opt(self):
             vlan_before = self.info['plugins'][0].get('vlan')
         except (IndexError, KeyError):
             vlan_before = None
-        vlan_after = self.params['opt'].get('vlan') if self.params['opt'] else None
+        vlan_after = self.params.get('opt', {}).get('vlan', None)
```
sshnaidm, line 389: It won't work that way. `self.params` are Ansible `params` which are always specified, and is `None` if not set. In `self.params.get('opt', {})` you'll always get `None` in case `opt` is not set in module args.
That's why it's implemented as implemented :) I can't understand how these changes are related to the issue, can you elaborate?

sshnaidm (changes requested):

plugins/modules/podman_network.py
sshnaidm: I see. Let me please fix it in a different place: #435

sshnaidm:
Tell me please if it solves your problem.

PR #442, opened 2022-06-27
podman-4.1.1 renames Healthcheck to Health

cjeanner:


raukadah:
LGTM, ship it :-)

JM1:
What is the reason for changing the name?

cjeanner:
> What is the reason for changing the name?

Well... no idea? podman folks renamed it, and it's creating an issue in tripleo ci here:
https://bugs.launchpad.net/tripleo/+bug/1979810
And we can't just go with my other proposed patch against tripleo-ansible, because our system will look for "healthcheck", won't catch it, and will then re-create the containers. Updating our code base is possible, but if anyone is having some custom things, it will blow to their face..

So it's best to add that alias here - also for the future.

BTW, I'll probably need some help on the doc related failures in the CI...

sshnaidm:
Podman renamed a keyword in status, but we don't follow status keywords, we follow the input parameters in the command line.
Instead of `healthcheck` we have now:
```bash
$ podman run --help | grep health
      --health-cmd string                        set a healthcheck command for the container ('none' disables the existing healthcheck)
```
So if we want to add alias, it should be `health_cmd` I think. Probably worth to add others like `s/healthcheck/health/`:
```
      --health-interval string                   set an interval for the healthchecks (a value of disable results in no automatic timer setup) (default "30s")
      --health-retries uint                      the number of retries allowed before a healthcheck is considered to be unhealthy (default 3)
      --health-start-period string               the initialization time needed for a container to bootstrap (default "0s")
      --health-timeout string                    the maximum time allowed to complete the healthcheck before an interval is considered failed (default "30s")
```

cjeanner:
Hmmmm so that one won't correct the tripleo issue. dang. I'll update the patch shortly to reflect @sshnaidm comment, but it's not that urgent anymore then.

sshnaidm:
> Hmmmm so that one won't correct the tripleo issue. dang. I'll update the patch shortly to reflect @sshnaidm comment, but it's not that urgent anymore then.

Right, it has nothing to do with the current issue in the job. Job vars should be adjusted.

cjeanner:
@sshnaidm apparently there was a bit more to do in order to support both names. And it may solve the issue on our side as well. WDYT?

cjeanner:
Closing - mostly useless at this point imho.

Issue #355, opened 2021-12-14
podman_container: mount breaks idempotency


smessmer:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When using the `mount` parameter for `podman_container`, the container gets recreated on every run even when there were no changes.

**Steps to reproduce the issue:**

The following runs fine and doesn't re-create the container if it already exists:

```
- name: POD registry CONTAINER registry
  containers.podman.podman_container:
    name: registry-registry
    image: registry:2
    pod: registry
    state: started
    network: registry-network
    volume:
    - /opt/containers/registry/config.v0.yml:/etc/docker/registry/config.yml
    - registry-data:/var/lib/registry
    - registry-log:/var/log
 ```

But the following re-creates the container on every run even when it already exists:

```
- name: POD registry CONTAINER registry
  containers.podman.podman_container:
    name: registry-registry
    image: registry:2
    pod: registry
    state: started
    network: registry-network
    mount:
    - "type=bind,source=/opt/containers/registry/config.v0.yml,destination=/etc/docker/registry/config.yml,ro=true"
    volume:
    - registry-data:/var/lib/registry
    - registry-log:/var/log
```

**Describe the results you received:**

**Describe the results you expected:**

**Additional information you deem important (e.g. issue happens only occasionally):**

**Version of the `containers.podman` collection:**
**Either git commit if installed from git: `git show --summary`**
**Or version from `ansible-galaxy` if installed from galaxy: `ansible-galaxy collection list | grep containers.podman`**

```
containers.podman             1.8.2  
```

**Output of `ansible --version`:**

```
ansible [core 2.12.1]
  config file = /home/heinzi/projects/dockerserver/host/ansible.cfg
  configured module search path = ['/home/heinzi/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/heinzi/projects/dockerserver/host/venv/lib/python3.9/site-packages/ansible
  ansible collection location = /home/heinzi/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/heinzi/projects/dockerserver/host/venv/bin/ansible
  python version = 3.9.5 (default, May 11 2021, 08:20:37) [GCC 10.3.0]
  jinja version = 3.0.3
  libyaml = True
```

**Output of `podman version`:**

```
podman version 3.0.1
```

**Output of `podman info --debug`:**

```
host:
  arch: amd64
[...]

PR #447, opened 2022-07-10
fix diffparam_volume and add diffparam_mount and diffparam_tmpfs

rubiksdot:
This is a fix for bug #355.

This compares the arguments podman recieved for the currently existing
container with the (effective) arguments to come.

This approach was taken over parsing Mounts from inspect because:

1. This line from https://github.com/containers/podman/blob/e084f0ee1e1ded564110e84eefca9f18b5669adf/libpod/container_inspect.go#L224
   regarding inspect's Mount value:

   "Only includes user-specified mounts. Only includes bind mounts and named volumes, not tmpfs volumes."

   Thus an incomplete solution would result.

2. The code required to parse so that it could be compared with the
   stuff to come was getting silly-level complex.

3. Anonymous volumes were impossible to decipher as both Name and
   Source are filled in with the podman-generated values.

Thus we compare the arguments podman create was run with to make the
existing container with the closest values to those arguments in the
new config.

This resulted in simpler code that takes care of the issue of anonymous
volumes.

The downside is that if someone moves, say, a tmpfs from mount to tmpfs
(or vice versa) that would reult in exactly the same result this will
be considered a different config. This can (possibly) be fixed if and
when it becomes an actual issue.

### DIFF
--- a/.github/workflows/collection-continuous-integration.yml
+++ b/.github/workflows/collection-continuous-integration.yml
@@ -13,9 +13,9 @@ jobs:
     strategy:
       matrix:
         runner-os:
-          - ubuntu-20.04
+          - ubuntu-22.04
         ansible-version:
-          - git+https://github.com/ansible/ansible.git@stable-2.11
+          - git+https://github.com/ansible/ansible.git@stable-2.13
         runner-python-version:
           - 3.8
     steps:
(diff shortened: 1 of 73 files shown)


rubiksdot:
sorry. total mess here. last time I push a commit whilst on the phone. :(

Issue #345, opened 2021-11-26
Rerunning container playbook with --new option in generate_systemd results in failure if the service is started


zephyros-dev:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Rerunning container playbook with --new option in generate_systemd results in failure if the service is started.

**Steps to reproduce the issue:**

1. Run a playbook with --new options in generate systemd and enable the systemd afterward
```
- hosts: localhost
  tasks:
    - name: HTTP Echo Container
      containers.podman.podman_container:
        name: "echo-test"
        image: "mendhak/http-https-echo:latest"
        ports:
          - "8080:80"
        generate_systemd:
          container_prefix: ""
          pod_prefix: ""
          separator: ""
          path: "/home/{{ ansible_user }}/.config/systemd/user"
          new: yes
        state: "started"

    - name: "Enable systemd service"
      ansible.builtin.systemd:
        daemon_reload: yes
        enabled: yes
        name: echo-test
        scope: user
        state: started
      register: service
```
2. The first run will work normally

**Describe the results you received:**
- The second run will consistently return the following error
```
fatal: [localhost]: FAILED! => {"changed": false, "msg": "Can't delete container echo-test", "stderr": "Error: no container with name or ID \"echo-test\" found: no such container\n", "stderr_lines": ["Error: no container with name or ID \"echo-test\" found: no such container"], "stdout": "", "stdout_lines": []}
```
**Describe the results you expected:**
- The second runs produce the container normally

**Additional information you deem important (e.g. issue happens only occasionally):**
- After the second run, no container can be found using podman ps -a
- The third run will consistently works exactly as the first run (state is changed, systemd file is created)
  - Due to this, a workaround can be made by using ansible retries. Example using the above playbook:
```
- hosts: localhost
  tasks:
    - name: HTTP Echo Container
      containers.podman.podman_container:
        name: "echo-test"
        image: "mendhak/http-https-echo:latest"
        ports:
          - "18080:80"
        generate_systemd:
          container_prefix: ""
          pod_prefix: ""
          separator: ""
          path: "/home/{{ ansible_user }}/.config/systemd/user"
          new: yes
        state: "started"
      # Retries until container run successfully
      register: container
      until: container.failed == false

    - name: "Enable systemd service"
      ansible.builtin.systemd:
        daemon_reload: yes
        enabled: yes
        name: echo-test
        scope: user
        state: started
      register: service
[...]

PR #453, opened 2022-07-14
Fix rerunning playbooks with generate_systemd --new

freundTech:
This fixes Issue #345 by making two changes:

1. If `generate_systemd.new` is set then `rm` is automatically set. This is needed because containers started by systemd services generated with the `--new` option always start with --rm. Otherwise ansible will try to restart container on every rerun, as ansible thinks `rm` should be off, while it has been turned on by the systemd service.
I also tried only changing `PodmanContainerDiff` for this, but that runs into the problem that then the container starts without `rm` when started by ansible but with `rm` once it has been restarted using systemd, leading to further problems.

2. Don't try to delete containers with `AutoRemove` set. As the containers are already removed by podman when they are stopped ansible will fail when it tries to remove them. This issue also affects containers with `rm` set manually, even when not using `generate_systemd`.

EDIT: Force-pushed to add sign-offs to commit

### DIFF
--- a/plugins/module_utils/podman/podman_container_lib.py
+++ b/plugins/module_utils/podman/podman_container_lib.py
@@ -1439,14 +1439,16 @@ class PodmanContainer:
         """Recreate the container."""
         if self.running:
             self.stop()
-        self.delete()
+        if not self.info['HostConfig']['AutoRemove']:
+            self.delete()
         self.create()
 
     def recreate_run(self):
         """Recreate and run the container."""
         if self.running:
             self.stop()
-        self.delete()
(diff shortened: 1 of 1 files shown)


sshnaidm:
LGTM, thanks, let me test it.

sshnaidm:
check-github

sshnaidm:
Perhaps it's worth to have a test about "--rm" and container deletion. Can you please think about it? For example here: https://github.com/sshnaidm/ansible-podman-collections/blob/master/tests/integration/targets/podman_container/tasks/main.yml

PR #455, opened 2022-07-19
New become plugin: podman_unshare

grzs:
## Summary
This is a basic become plugin to leverage podman unshare capabilities.

## Examples
```
- name: create file
  become_method: containers.podman.podman_unshare
  become: yes
  ansible.builtin.file:
    path: /tmp/test.file
    owner: 1000
    group: 500
- name: show files
  become_method: containers.podman.podman_unshare
  become: yes
  ansible.builtin.command: ls -ln /tmp/
```


sshnaidm:
I think it's very cool! Probably worth to have some examples and tests. Maybe also to put a link to podman unshare docs?

grzs:
@sshnaidm thank you for the reply! My English is not the best, but I try to do my best and put together some examples and docs, and write tests too.

sshnaidm:
> @sshnaidm thank you for the reply! My English is not the best, but I try to do my best and put together some examples and docs, and write tests too.

No worries about English! I'll help you to fix it. Thanks for this idea!

sshnaidm:
@grzs I think we can start from adding a few examples inside, just to present a few use cases,

plugins/become/podman_unshare.py
```diff
@@ -0,0 +1,34 @@
...
+
+DOCUMENTATION = """
+    name: podman_unshare
+    short_description: Run tasks using podman unshare
+    description:
+        - This become plugins allows your remote/login user to execute commands in its container user namespace
+    author:
+    - Janos Gerzson (@grzs)
+    version_added: 1.9.0
+"""
```
sshnaidm, line 16: Like adding here "EXAMPLES" section as we have in modules.

grzs: @sshnaidm  Sorry, I am in a tight sprint. Yes, I will add the EXAMPLES section hopefully today.

grzs: @sshnaidm I added something, don't know if it is what you expected. I also inserted a link to the podman documentation, as you mentioned. I am not sure if it is the right format though.
Are unit tests needed as well? Do you have an idea for a test case?

grzs: There is a limitation which can be very inconvenient: this plugin is useful only when one has access to the user what she/he wants to operate with. Because its a become plugin and such plugin cannot be chained, one have to choose between this and sudo or so. A solution could be another plugin, like podman_unshare_sudo, what would be basically the standard sudo plugin powered up with podman unshare.
(eg. original command 'id' transformed into 'sudo -u foobar podman unshare id'). What do you think?

sshnaidm: Maybe we can have additional option to become plugin, which will be used for setting a user?
Like `become_user: <username>`? Become plugins can have multiple options, we can use it, see for example: https://github.com/ansible-collections/community.general/blob/main/plugins/become/doas.py#L15-L80

grzs: Yes, you are right. I was considering that option but wanted to keep it simple, and as become_user is irrelevant regarding podman unshare, I omitted it. But I can make this option available if you think it's better to have one unified plugin.

plugins/become/podman_unshare.py
```diff
@@ -0,0 +1,91 @@
...
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+DOCUMENTATION = """
+    name: podman_unshare
+    short_description: Run tasks using podman unshare
+    description:
+        - This become plugins allows your remote/login user
+          to execute commands in its container user namespace.
+          Official documentation: https://docs.podman.io/en/latest/markdown/podman-unshare.1.html
```
sshnaidm, line 14: The colon in this line causes the error:

ERROR: plugins/become/podman_unshare.py:14:33: error: DOCUMENTATION: syntax error: mapping values are not allowed here (syntax)

grzs: I see, thanks

plugins/become/podman_unshare.py
grzs: @sshnaidm Hi, what should I add to make this PR complete? Do we need unit tests? Should I add the become_user option via sudo?
Thanks

sshnaidm: @grzs hey, I think we can skip unit tests for now, I can handle that later. Just to finish with become_user as we discussed above, and I think we can merge it. Thoughts?

grzs: great! that looks simple. I would assemble something like this:
sudo -u <become_user> -i podman unshare <command>

The standard sudo plugin is more complicated,  I don't know why. I tested the above command on debian. I think we can narrow the usage to linux only as podman lives on linux solely as far as I understood.
You can check the sudo plugin here: https://github.com/ansible-collections/community.general/blob/main/plugins/become/sudosu.py

I can push a draft tomorrow.

grzs: Hi @sshnaidm, it was a long day since 'yesterday', sorry for that. I wasn't sure about the implementation and wanted to test some options beforehand, but my linux boxes with podman was all down, but finally got to deliver a solution. Mostly I used parts from the builtin sudo solution (https://github.com/ansible/ansible/blob/devel/lib/ansible/plugins/become/sudo.py) instead of community.sudosu, and omitted the flags option because I found out that when I tried to run podman unshare without the login shell option of sudo, it couldn't chdir out from root directory. So there is no flags option, but possible to customize the password prompt like it is with the sudo plugin. Things getting complicated only when I want to use $HOME or '~' in the command because they still refer to the ansible_user home directory. It could be avoided with the sudo -H flag baut that can't be use alongside with -i which is mandatory for podman unshare. But I think it's possible to live with it. Here is an example playbook I tested with:
```
---
- name: unshare test
  gather_facts: no
  hosts: debian_unshare_test
  vars:
    pasas_home: /home/pasas
    ansible_become_user: pasas
    # ansible_become_method: containers.podman.podman_unshare
  tasks:
  - name: dir exists
    become: yes
    file:
      state: directory
      path: ~/unshare
  - name: dir exists
    become: yes. # sudo is the default become plugin
    file:
      state: directory
      path: ~/unshare. # '~' works as expacted
  - name: file exists
    become_method: containers.podman.podman_unshare
    become: yes
    copy:
      content: bump
      dest: "{{ pasas_home }}/unshare/y"
      owner: 1000
      group: 500
  - name: show files
    become: yes
    become_method: sudo. # unnecessery
    command: "ls -ln {{ pasas_home }}/unshare/"
    register: result
  - debug:
      var: result.stdout
```

sshnaidm: Thanks @grzs ! Sorry for delay.

grzs: no problem, and thank you for your assistance and acceptance!

PR #458, opened 2022-07-23
podman_container should ensure image by using os path if rootfs is used

mainnika:
if the `rootfs` is true then podman_container should use an image argument as a path and check the existence by using os-path method. 
if there is no such a directory then the module should fire an error.

playbook example
```
- name: Container present
    name: mycontainer
    state: present
    rootfs: yes
    image: /opt/containers/mycontainer-rootfs
    command: "sleep 1d"
```

↑ this currently fails with
```
{"changed": false, "msg": "Can't pull image /opt/containers/mycontainer-rootfs", "stderr": "Error: invalid reference format\n", "stderr_lines": ["Error: invalid reference format"], "stdout": "", "stdout_lines": []}
```


sshnaidm:
As I see, the test ansible-podman-collections/tests/integration/targets/podman_container/tasks/main.yml:426 fails:
https://github.com/containers/ansible-podman-collections/blob/99c2851341999bfeb0b78458efa09f874c923a9e/tests/integration/targets/podman_container/tasks/main.yml#L419-L437

```diff
 --- before
+++ after
@@ -1 +1 @@
-label - {}
+label - None
```

mainnika:
> As I see, the test ansible-podman-collections/tests/integration/targets/podman_container/tasks/main.yml:426 fails:
> ```diff
>  --- before
> +++ after
> @@ -1 +1 @@
> -label - {}
> +label - None
> ```

Yes, I see, changed this line back. `or` works here better than a default value.

sshnaidm:
The failure in job is not relater, I fixed it in https://github.com/containers/ansible-podman-collections/commit/e6544bdd0a66c558e3e66caf97aae75b75720b33, so you can rebase on latest master.

sshnaidm:
@mainnika would you like to add a test to https://github.com/containers/ansible-podman-collections/blob/master/tests/integration/targets/podman_container/tasks/main.yml ?

sshnaidm:
check-github

mainnika:
@sshnaidm 
Ok, will do that

sshnaidm:
check-github

sshnaidm:
@mainnika I can merge and you can add tests later, but I just try to keep it tested always :)

mainnika:
I also like when everything is tested. I'll try to find a time either today or tomorrow morning.

mainnika:
@sshnaidm
Added tests, could you please take a look, maybe I missed some cases.

sshnaidm:
Awesome! Thanks for your great contribution!

PR #461, opened 2022-08-08
Add required argument to example

weichweich:
The `podman_play` example doesn't work because a required argument is missing.


weichweich (commented):
.

sshnaidm:
Hmm, probably worth to make this state the default one. Seems like I overlooked it.

Issue #491, opened 2022-10-20
generate_systemd don't work for a pod


CyberFox001:
**Is this a BUG REPORT or FEATURE REQUEST?**

/kind bug

**Description**

If I generate a systemd .service file for a pod, when I create the pod
before the containers creation, then systemd fail to start the
service.

I got this message:
```
Unable to start service pod-peertube_instance.service: Job for pod-peertube_instance.service failed because the service did not take the steps required by its unit configuration.
See "systemctl --user status pod-peertube_instance.service"
```

I think it's because the .service file is generated before the add of
containers to the pod. So it miss the services of the containers in
its dependencies.

If I generate the systemd .service file for this pod after the creation of the containers, in a separate task like this:
``` yaml
- name: Generate the .service file for the pod
  containers.podman.podman_pod:
    name: "{{ peertube_pod_name }}"
    generate_systemd:
      path: "{{ systemd_services_path }}"
      restart_policy: always
      names: true
      container_prefix: peertube
```

The result is the deletion of the all the containers inside the pod, except for the infra container.

In generale, the documentation about the systemd integration with pod or containers is unclear:
* Does the generation of systemd file from a pod also generate the .service file of its containers ?
* When do I generate the systemd files ? At the pod creation ? After the add of containers in the pod ?
* Do I have to stop the container befor it's integration in systemd ?

**Steps to reproduce the issue:**

1. On your playbook, have a list of tasks that create a pod with a few containers and generate the systemd files in the same time, like this:
``` yaml
# Pod creation
- name: Be sure the pod exist
  containers.podman.podman_pod:
    name: "{{ peertube_pod_name }}"
    state: created
    ports:
      - 1935:1935
      - 9000:9000
    generate_systemd:
      path: "{{ systemd_services_path }}"
      restart_policy: always
      names: true
      container_prefix: peertube
      
# Container creation
- name: Be sure the Postgresql container exist
  containers.podman.podman_container:
    pod: "{{ peertube_pod_name }}"
    name: "{{ postgres_container_name }}"
    image: "{{ postgres_image_path }}:{{ postgres_version }}"
    state: stopped
    env:
      POSTGRES_DB: "{{ postgres_database_name }}"
      POSTGRES_USER: "{{ postgres_user }}"
      POSTGRES_PASSWORD: "{{ postgres_password }}"
    volume:
      - "{{ postgres_volume_data_name }}:/var/lib/postgresql/data"
    generate_systemd:
[...]

PR #498, opened 2022-10-25
Podman generate systemd

CyberFox001:
This pull request add a new module: `podman_generate_systemd`.

This module will generate systemd .service units for pods and containers.

In this Pull Request, I provide:
- The module code
- It's documentation
- An integration test

The module can be used to only generate the systemd .service units and
save them in an Ansible variable, or to create the unit files on the
host. It support the check mode.

If you specify a Pod, le unit for the pod and all its containers will
be generated.

The module support a lot of options:
- `name`: Name of the pod or container to export as systemd unit
- `dest`: Destination of the generated systemd unit file(s)
- `new`: Make the unit create the container before starting it
- `restart_policy`: Restart policy of the service
- `restart_sec`: Configures the time to sleep before restarting a service (as configured with restart-policy)
- `start_timeout`: Override the default start timeout for the container with the given value in seconds
- `stop_timeout`: Override the default stop timeout for the container with the given value in seconds
- `env`: Set environment variables to the systemd unit files, need Podman >= 4.3.0 to be used
- `use_name`: Use name of the containers for the start, stop, and description in the unit file
- `container_prefix`: Set the systemd unit name prefix for containers
- `pod_prefix`: Set the systemd unit name prefix for pods
- `separator`: Systemd unit name separator between the name/id of a container/pod and the prefix
- `no_header`: Do not generate the header including meta data such as the Podman version and the timestamp
- `after`: Add a systemd unit in the `after` option of the generated systemd unit, need Podman >= 4.0.0 to be used
- `wants`: Add a systemd unit in the `wants` option of the generated systemd unit, need Podman >= 4.0.0 to be used
- `requires`: Add a systemd unit in the `requires` option of the generated systemd unit, need Podman >= 4.0.0 to be used
- `executable`: Podman executable name or full path

In the return values, you get the generated systemd units and the
command who was run to generate thems.

I write my own function to generate the systemd units, not the already
existed function. But they are similar.

### DIFF
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -319,7 +319,7 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str], str]:
             f' Command returned with code: {return_code}.'
             f' Error message: {stderr}.'
         )
-        return changed, {}
+        return changed, {}, f'{command}'
 
     # In case of command execution success, its stdout is a json
     # dictionary. This dictionary is all the generated systemd units.
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -277,35 +277,21 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str], str]:
     #  After option (only for Podman 4.0.0 and above)
     after = module.params.get('after')
     if after:
-        # If after is a single string
-        if isinstance(after, str):
-            command_options.append(f'--after={after}')
-        # If wants is a list
-        if isinstance(after, list):
-            for item in after:
-                command_options.append(f'--after={item}')
+        for item in after:
+            command_options.append(f'--after={item}')
 
     #  Wants option (only for Podman 4.0.0 and above)
     wants = module.params.get('wants')
     if wants:
-        # If wants is a single string
-        if isinstance(wants, str):
-            command_options.append(f'--wants={wants}')
-        # If wants is a list
-        if isinstance(wants, list):
-            for item in wants:
-                command_options.append(f'--wants={item}')
+        for item in wants:
+            command_options.append(f'--wants={item}')
                 
     #  Requires option (only for Podman 4.0.0 and above)
     requires = module.params.get('requires')
     if requires:
-        # If requires is a single string
-        if isinstance(requires, str):
-            command_options.append(f'--requires={requires}')
-        # If requires is a list
-        if isinstance(requires, list):
-            for item in requires:
-                command_options.append(f'--requires={item}')
+        for item in requires:
+            command_options.append(f'--requires={item}')
+
     
     #  Full command, with option include
     command_options.extend(['--format', 'json'])
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -223,6 +223,112 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str]]:
             )
     # Return the systemd .service unit(s) content
     return changed, systemd_units
+
+def run_module():
+    '''Run the module on the target'''
+    # Build the list of parameters user can use
+    module_parameters = {
+        'name': {
+            'type': 'str',
+            'required': True,
+        },
+        'dest': {
+            'type': 'str',
+            'required': False,
+        },
+        'new': {
+            'type': 'bool',
+            'required': False,
+            'default': False,
+        },
+        'restart_policy': {
+            'type': 'str',
+            'required': False,
+            'default': 'on-failure',
+            'choices': RESTART_POLICY_CHOICES,
+        },
+        'restart_sec': {
+            'type': 'int',
+            'required': False,
+        },
+        'start_timeout': {
+            'type': 'int',
+            'required': False,
+        },
+        'stop_timeout': {
+            'type': 'int',
+            'required': False,
+        },
+        'use_names': {
+            'type': 'bool',
+            'required': False,
+            'default': True,
+        },
+        'container_prefix': {
+            'type': 'str',
+            'required': False,
+        },
+        'pod_prefix': {
+            'type': 'str',
+            'required': False,
+        },
+        'separator': {
+            'type': 'str',
+            'required': False,
+        },
+        'no_header': {
+            'type': 'bool',
+            'required': False,
+            'default': False,
+        },
+        'after': {
+            'type': 'list',
+            'elements': 'str',
+            'required': False,
+        },
+        'wants': {
+            'type': 'list',
+            'elements': 'str',
+            'required': False,
+        },
+        'requires': {
+            'type': 'list',
+            'elements': 'str',
+            'required': False,
+        },
+        'debug': {
+            'type': 'bool',
+            'required': False,
+            'default': False,
+        },
+        'executable': {
+            'type': 'str',
+            'required': False,
+            'default': 'podman',
+        },
+    }
+    
+    # Build result dictionary
+    result = {
+        'changed': False,
+        'systemd_units': {},
+    }
+
+    # Build the Ansible Module
+    module = AnsibleModule(
+        argument_spec=module_parameters,
+        supports_check_mode=True
+    )
+
+    # Generate the systemd units
+    state_changed, systemd_units = generate_systemd(module)
+
+    result['changed'] = state_changed
+    result['systemd_units'] = systemd_units
+
+    # Return the result
+    module.exit_json(**result)
+
 def main():
     pass
 
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -319,7 +319,6 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str], str]:
             f' Command returned with code: {return_code}.'
             f' Error message: {stderr}.'
         )
-        return changed, {}, f'{command}'
 
     # In case of command execution success, its stdout is a json
     # dictionary. This dictionary is all the generated systemd units.
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -221,6 +221,8 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str]]:
                 'Error writing systemd units files: '
                 f'{exception}'
             )
+    # Return the systemd .service unit(s) content
+    return changed, systemd_units
 def main():
     pass
 
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -31,6 +31,7 @@ options:
       - Refer to podman-generate-systemd(1) man page for more information.
     type: bool
     required: false
+    default: false
   restart_policy:
     description:
       - Restart policy of the service
@@ -69,7 +70,7 @@ options:
       - Only with Podman 4.3.0 and above
     type: dict
     required: false
-  use_name:
+  use_names:
     description:
       - Use name of the containers for the start, stop, and description in the unit file.
     type: bool
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -234,14 +234,14 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str], str]:
     '''
     # Flag which indicate whether the targeted system state is modified
     changed = False
-    
+
     # Build the podman command, based on the module parameters
     command_options = []
 
     #  New option
     if module.params.get('new'):
         command_options.append('--new')
-    
+
     #  Restart policy option
     restart_policy = module.params.get('restart_policy')
     if restart_policy:
@@ -304,7 +304,7 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str], str]:
     if wants:
         for item in wants:
             command_options.append(f'--wants={item}')
-                
+
     #  Requires option (only for Podman 4.0.0 and above)
     requires = module.params.get('requires')
     if requires:
@@ -318,7 +318,7 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str], str]:
             command_options.append(
                 f"-e='{env_var_name}={env_var_value}'",
             )
-    
+
     #  Full command, with option include
     command_options.extend(['--format', 'json'])
     command = [
@@ -327,7 +327,7 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str], str]:
         module.params['name'],
     ]
     command_str = ' '.join(command)
-    
+
     # Run the podman command to generated systemd .service unit(s) content
     return_code, stdout, stderr = module.run_command(command)
 
@@ -348,7 +348,7 @@ def generate_systemd(module: AnsibleModule) -> tuple[bool, list[str], str]:
 
     # Load the returned json dictionary as a python dictionary
     systemd_units = json.loads(stdout)
-    
+
     # Write the systemd .service unit(s) content to file(s), if
     # requested and not in check mode
     if module.params.get('dest') and not module.check_mode:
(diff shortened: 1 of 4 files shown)


CyberFox001:
So, the first sanity check have 2 problems: My name and a TypeError.

For the first one, is it because of the accent in my first name ?

For the TypeError in `argument_spec`, is it because I use a more classic way to define the dictionaries ?

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+options:
+  name:
+    description:
+      - Name of the pod or container to export
+    type: str
+    required: true
+  dest:
+    description:
+      - Destination of the generated systemd unit file(s)
+    type: str
```
sshnaidm, line 26: `ERROR: plugins/modules/podman_generate_systemd.py:0:0: doc-type-does-not-match-spec: Argument 'dest' in argument_spec defines type as 'path' but documentation defines type as 'str'`

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+        command_options.append('--new')
+
+    #  Restart policy option
+    restart_policy = module.params.get('restart_policy')
+    if restart_policy:
+        # If the restart policy requested is not supported by podman
+        if restart_policy not in RESTART_POLICY_CHOICES:
+            # Then stop the module execution and return an error message
+            module.fail_json(
+                msg=f'Restart policy requested is "{restart_policy}"'
```
sshnaidm, line 253: `ERROR: plugins/modules/podman_generate_systemd.py:253:69: traceback: SyntaxError: invalid syntax`
Can we please make it Py2 compatible?

sshnaidm (changes requested):
A few comments from failing jobs.

plugins/modules/podman_generate_systemd.py
CyberFox001: Ah, I forget to update the documentation.

Done.

plugins/modules/podman_generate_systemd.py
CyberFox001: Do we target Python 2 ? I thought that we targeted Python 3.6.8. [from](https://github.com/containers/ansible-podman-collections/issues/491#issuecomment-1287773280)

sshnaidm: Sorry, probably I forgot by myself :laughing: But it's in todo list to drop Py2 support, since it requires to release the major collection version (2.0.0), which will be included in next major Ansible release only (because it's not backward compatible). So I'd like to collect all non-backward-compatible changes and release it later.

CyberFox001: Do I made modification to be Python 2 compatible ?

sshnaidm: Yeah, just use `.format` or `%s`. Seems like it's only the place?

sshnaidm: Would be great if you include this code as well: https://github.com/containers/ansible-podman-collections/pull/501 just copy paste it to your patch. That way we'll run CI jobs here.

CyberFox001: > Yeah, just use `.format` or `%s`. Seems like it's only the place?

I use a lot of f'' string.

CyberFox001: > Would be great if you include this code as well: #501 just copy paste it to your patch. That way we'll run CI jobs here.

Sorry, I did not understand what you meant.

sshnaidm: Code in https://github.com/containers/ansible-podman-collections/pull/501 is a skeleton for tests to run them in CI. Can you please take this code to your current patch?

CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+            if not os.path.exists(systemd_units_dest):
+                # Make it
+                os.makedirs(systemd_units_dest)
+                changed = True
+            # If destination exist but not a directory
+            if not os.path.isdir(systemd_units_dest):
+                # Stop and tell user that the destination is not a directry
+                module.fail_json(
+                    f'Destination {systemd_units_dest} is not a directory.'
+                    "Can't save systemd unit files in."
```
sshnaidm: It's better to use same quotes for messages (either single or double)

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+    ]
+    command_str = ' '.join(command)
+
+    # Run the podman command to generated systemd .service unit(s) content
+    return_code, stdout, stderr = module.run_command(command)
+
+    # In case of error in running the command
+    if return_code != 0:
+        # Print informations about the error and return and empty dictionary
+        module.fail_json(
```
sshnaidm, line 343: In jobs it fails here:
```
File "/opt/hostedtoolcache/Python/3.9.15/x64/lib/python3.9/runpy.py", line 97, in _run_module_code/podman_generate_systemd.py", line 509, in run_module

...

  File "/tmp/ansible_containers.podman.podman_generate_systemd_payload_oeq4snf7/ansible_containers.podman.podman_generate_systemd_payload.zip/ansible_collections/containers/podman/plugins/modules/podman_generate_systemd.py", line 340, in generate_systemd
TypeError: fail_json() takes 1 positional argument but 2 were given
fatal: [localhost]: FAILED! => {
    "changed": false,
    "rc": 1
}
```

plugins/modules/podman_generate_systemd.py
CyberFox001: Finally, what do I do about Python 2 ?

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+    # dictionary. This dictionary is all the generated systemd units.
+    # Each key value pair is one systemd unit. The key is the unit name
+    # and the value is the unit content.
+
+    # Load the returned json dictionary as a python dictionary
+    systemd_units = json.loads(stdout)
+
+    # Write the systemd .service unit(s) content to file(s), if
+    # requested and not in check mode
+    if module.params.get('dest') and not module.check_mode:
```
sshnaidm, line 363: Is `check_mode` processed somewhere?

sshnaidm (commented):
A few comments inside

plugins/modules/podman_generate_systemd.py
sshnaidm: We need to convert all F-strings to `format`. Let me know if you need help with that, I can do it.

plugins/modules/podman_generate_systemd.py
sshnaidm: Ok, it's fixed already.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+# 2022, Sébastien Gendre <seb@k-7.ch>
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+from __future__ import (absolute_import, division, print_function)
+__metaclass__ = type
+
+
+DOCUMENTATION = '''
+module: podman_generate_systemd
+author:
+  - Sébastien Gendre (@CyberFox001)
```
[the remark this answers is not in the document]
CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+      - Keys are the environment variable names, and values are the environment variable values
+      - Only with Podman 4.3.0 and above
+    type: dict
+    required: false
+  use_names:
+    description:
+      - Use name of the containers for the start, stop, and description in the unit file.
+    type: bool
+    required: false
+    default: yes
```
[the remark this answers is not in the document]
CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+      - If not set, use the default defined by podman, C(-).
+      - Refer to podman-generate-systemd(1) man page for more information.
+    type: str
+    required: false
+  no_header:
+    description:
+      - Do not generate the header including meta data such as the Podman version and the timestamp.
+    type: bool
+    required: false
+    default: no
```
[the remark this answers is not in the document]
CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+      - Only with Podman 4.0.0 and above
+    type: list
+    elements: str
+    required: false
+  executable:
+    description:
+      - C(Podman) executable name or full path
+    type: str
+    required: false
+    default: podman
```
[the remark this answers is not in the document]
CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+
+    #  New option
+    if module.params.get('new'):
+        command_options.append('--new')
+
+    #  Restart policy option
+    restart_policy = module.params.get('restart_policy')
+    if restart_policy:
+        # If the restart policy requested is not supported by podman
+        if restart_policy not in RESTART_POLICY_CHOICES:
```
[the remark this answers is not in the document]
CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+                msg=f'Restart policy requested is "{restart_policy}"'
+                f',  but must be one of: {RESTART_POLICY_CHOICES}'
+            )
+        # Else add the restart policy to options
+        if restart_policy == 'no-restart':
+            restart_policy = 'no'
+        command_options.append(f'--restart-policy={restart_policy}')
+
+    #  Restart-sec option (only for Podman 4.0.0 and above)
+    restart_sec = module.params.get('restart_sec')
```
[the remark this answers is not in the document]
CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+            )
+
+    #  Full command, with option include
+    command_options.extend(['--format', 'json'])
+    command = [
+        module.params.get('executable'), 'generate', 'systemd',
+        *command_options,
+        module.params['name'],
+    ]
+    command_str = ' '.join(command)
```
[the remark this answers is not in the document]
CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
CyberFox001: `check_mode` is just "run the module but don't do any modification on the target host" ? Or did I miss something ?

plugins/modules/podman_generate_systemd.py
CyberFox001: It's ok, I've done it.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,525 @@
...
+      - Destination of the generated systemd unit file(s)
+    type: path
+    required: false
+  new:
+    description:
+      - Generate unit files that create containers and pods, not only start them.
+      - Refer to podman-generate-systemd(1) man page for more information.
+    type: bool
+    required: false
+    default: false
```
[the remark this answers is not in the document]
CyberFox001: Done.

plugins/modules/podman_generate_systemd.py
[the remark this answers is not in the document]
CyberFox001: So, using `shlex.join()` generate a weird command with the `-e='ENV_VAR=value` option. And it's seems not possible to avoid it.

As `command_str` is only used for module return and it's made with user given options, maybe I can come back to `' '.join(command)`

sshnaidm:
Just trailing whitespaces to remove and rebase on latest master, and you're all set.
```
ERROR: plugins/modules/podman_generate_systemd.py:349:47: trailing-whitespace: Trailing whitespace
ERROR: plugins/modules/podman_generate_systemd.py:360:0: trailing-whitespace: Trailing whitespace
```

plugins/modules/podman_generate_systemd.py
sshnaidm: In general it should report "changed" is something **would be** changed, but w/o actually doing anything. And to report "ok" when nothing should be changed.

plugins/modules/podman_generate_systemd.py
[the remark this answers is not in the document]
sshnaidm: Ack

plugins/modules/podman_generate_systemd.py
sshnaidm: If you want to change it you can do it in the next patch.

CyberFox001: Done.

CyberFox001:
> Just trailing whitespaces to remove and rebase on latest master, and you're all set.
> 
> ```
> ERROR: plugins/modules/podman_generate_systemd.py:349:47: trailing-whitespace: Trailing whitespace
> ERROR: plugins/modules/podman_generate_systemd.py:360:0: trailing-whitespace: Trailing whitespace
> ```

Done.

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,572 @@
...
+        message += ' Error message: {stderr}.'
+        module.fail_json(
+            msg=message.format(
+                command_str=command_str,
+                return_code=return_code,
+                stderr=stderr,
+            ),
+            changed=changed,
+            systemd_units={},
+            podman_command=command_str,
```
sshnaidm, line 379: This should be removed, `fail_json` accepts only one arg `msg`

plugins/modules/podman_generate_systemd.py
```diff
@@ -0,0 +1,572 @@
...
+                # Stop and tell user that the destination is not a directry
+                message = "Destination {systemd_units_dest} is not a directory."
+                message += " Can't save systemd unit files in."
+                module.fail_json(
+                    msg=message.format(
+                        systemd_units_dest=systemd_units_dest,
+                    ),
+                    changed=changed,
+                    systemd_units=systemd_units,
+                    podman_command=command_str,
```
sshnaidm, line 412: Ditto

sshnaidm (changes requested):
Sorry, but noticed the problem with `fail_json` in the current version.

plugins/modules/podman_generate_systemd.py
CyberFox001: I checked the source code of `fail_json()`: [Link](https://github.com/ansible/ansible/blob/devel/lib/ansible/module_utils/basic.py#L1516)

This method accept other arguments and will return them as a JSON dictionary at the output of the module. Like `exit_json()` will do.

You can found un example here: [systemd_service module](https://github.com/ansible/ansible/blob/devel/lib/ansible/modules/systemd_service.py#L563)

Returning `changed`, `systemd_units` and `podman_command` is necessary because you can call an Ansible module with an option to ignore errors and store the output data in an Ansible variable.
For this case, `podman_generate_systemd` module need to return these 3 values so users can use them.

plugins/modules/podman_generate_systemd.py
CyberFox001: Ditto

plugins/modules/podman_generate_systemd.py
sshnaidm: Great, thanks for clarification!

sshnaidm (approved):

sshnaidm:
Thanks for awesome patch!

CyberFox001:
Thank you for your help, and patience. :)

Issue #428, opened 2022-05-19
Warning about improperly configured remote target


nkakouros:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

When running with latest ansible (devel branch), a warning shows on every task run using the podman connection plugin. I am using molecule

**Steps to reproduce the issue:**

1. Will provide if indeed needed

**Describe the results you received:**

```
TASK [twmn.twmn.ubuntu_setup : Gather facts] ***********************************
[WARNING]: The "podman" connection plugin has an improperly configured remote
target value, forcing "inventory_hostname" templated value instead of the
string
ok: [molecule_test]
```

**Describe the results you expected:**
No warnings

**Additional information you deem important (e.g. issue happens only occasionally):**
[Here](https://github.com/ansible/ansible/issues/77001) is the ansible issue where this bug was discussed for the docker collection.

**Version of the `containers.podman` collection:**
**Either git commit if installed from git: `git show --summary`**
**Or version from `ansible-galaxy` if installed from galaxy: `ansible-galaxy collection list | grep containers.podman`**

```
1.9.3 (installed by molecule)
```

**Output of `ansible --version`:**

```
ansible [core 2.14.0.dev0]
  config file = None
  configured module search path = ['/home/nikos/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /usr/lib/python3.10/site-packages/ansible
  ansible collection location = /home/nikos/.ansible/collections:/usr/share/ansible/collections
  executable location = /usr/bin/ansible
  python version = 3.10.4 (main, Mar 23 2022, 23:05:40) [GCC 11.2.0] (/usr/bin/python3.10)
  jinja version = 3.1.2
  libyaml = True
```

**Output of `podman version`:**

```
Client:       Podman Engine
Version:      4.1.0
API Version:  4.1.0
Go Version:   go1.18.1
Git Commit:   e4b03902052294d4f342a185bb54702ed5bed8b1
Built:        Fri May  6 18:18:30 2022
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

``` yaml
host:
  arch: amd64
  buildahVersion: 1.26.1
  cgroupControllers:
  - memory
  - pids
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: /usr/bin/conmon is owned by conmon 1:2.1.0-1
    path: /usr/bin/conmon
[...]

PR #506, opened 2022-11-01
Warning about improperly configured remote target

rpelisse:
This is a very naive implementation of the fix suggested on the issue #428. Please review and comment!

### DIFF
--- a/plugins/connection/podman.py
+++ b/plugins/connection/podman.py
@@ -23,6 +23,8 @@ DOCUMENTATION = '''
         default: inventory_hostname
         vars:
           - name: ansible_host
+          - name: inventory_hostname
+          - name: ansible_podman_host
       remote_user:
         description:
             - User specified via name or UID which is used to execute commands inside the container. If you


plugins/connection/podman.py
```diff
@@ -82,10 +82,17 @@ class Connection(ConnectionBase):
...
+      # compat
+      if _container_id is None and self._play_context.remote_addr is not None:
+        _container_id = self._play_context.remote_addr
+      return _container_id
+
     def __init__(self, play_context, new_stdin, *args, **kwargs):
         super(Connection, self).__init__(play_context, new_stdin, *args, **kwargs)
 
-        self._container_id = self._play_context.remote_addr
+        self._container_id = self._get_container_id
```
sshnaidm, line 95: `self._get_container_id()` ?

rpelisse: @sshnaidm thanks!

rpelisse:
```
FATAL: The 1 sanity test(s) listed below (out of 32) failed. See error output above for details.
pep8
```
OK, so there is some [formatting issue](https://peps.python.org/pep-0008/) in my code change. I'll try to fix that asap.

plugins/connection/podman.py
```diff
@@ -82,10 +82,17 @@ class Connection(ConnectionBase):
     # users will start containers and fail to connect to them.
     has_pipelining = False
 
+    def _get_container_id(self):
+        _container_id = self.get_option('remote_addr')
+        # compat
+        if _container_id is None and self._play_context.remote_addr is not None:
+          _container_id = self._play_context.remote_addr
```
ag0783, line 89:
```suggestion
            _container_id = self._play_context.remote_addr
```

rpelisse:
@ag0783 Thanks! I was looking into those formatting issue last week, then I got distracted and never came back to it...

sshnaidm:
This fails with error `The error was: AttributeError: 'Connection' object has no attribute '_load_name'`

rpelisse:
Yep, I saw that. I need to take a look at Python API (as you may have guessed, I'm far from being a python developer) and see what's wrong here.

adrianlzt:
Just adding these two lines fixes the bug
```
--- podman.py.orig      2022-11-08 06:56:09.427326681 +0100
+++ podman.py.new       2022-11-08 07:51:16.936084988 +0100
@@ -23,6 +23,8 @@
         default: inventory_hostname
         vars:
           - name: ansible_host
+          - name: inventory_hostname
+          - name: ansible_podman_host
       remote_user:
         description:
             - User specified via name or UID which is used to execute commands inside the container. If you
```

In fact, just with the ``inventory_hostname`` it is fixed, but I added also de ansible_podman_host to do as in the [docker plugin](https://github.com/bcoca/community.docker/blob/82df7291e22ed3a9b5cdb78c39158d86006b82b3/plugins/connection/docker.py#L32).

I don't fully understand why this works though.

m-he:
> 

it works, thanks a lot!

rpelisse:
Awesome! I'll update the PR today, hopefully, it will finally passes and be ready for merging. Thanks for all the help guys.

adrianlzt:
You can remove the ``self._get_container_id()`` also.

plugins/connection/podman.py
```diff
@@ -82,10 +84,15 @@ class Connection(ConnectionBase):
     # users will start containers and fail to connect to them.
     has_pipelining = False
 
+    def _get_container_id(self):
+        _container_id = self.get_option('remote_addr')
+        if _container_id is None and self._play_context.remote_addr is not None:
+            _container_id = self._play_context.remote_addr
+        return _container_id
```
sshnaidm, line 91: please remove it if you don't use this function

sshnaidm (changes requested):

sshnaidm:
Please sign your commit with `git commit -s` to satisfy DCO job.
You can do `git commit --amend -s` and then push force.

plugins/connection/podman.py
sshnaidm: Didn't you want to remove this function and `_container_id` completely?

rpelisse: That was the suggestion above, but if I do this there is literally no code changes in the PR (so I suspects it won't fix the problem). 

More importantly, the tests failed complaining about the missing _container_id. So I put both back, to see if the tests passes this time.

sshnaidm: If I understood @adrianlzt correctly, you don't need to change a code, but just to add these lines:
```
+          - name: inventory_hostname
+          - name: ansible_podman_host
```

sshnaidm (approved):

sshnaidm:
Probably same thing as in https://github.com/ansible/ansible/issues/77841#issuecomment-1130211123

sshnaidm:
Thanks for your contribution!

abalage:
I ran into this yesterday and I tested the proposed patch manually and it worked for me.
I see the MR is closed now, so all I want to say is thank you.

olgeni:
The fix with "inventory_hostname" will also improve the situation with the jail connection plugin 👀

sshnaidm:
@olgeni you might want to add it here: https://github.com/ansible-collections/community.general/blob/d2094669859fe4b48aa20f1b3ae868b61023b51d/plugins/connection/jail.py#L24-L26

olgeni:
Yup it works ✌️

PR #517, opened 2022-11-21
feat: allow empty string for prefixes

### LINKED ISSUE #516: Allow empty string for container_prefix, pod_prefix and separator in `podman_generate_systemd` module

/kind feature

**Description**

The podman generate systemd command allow using empty string in its parameter. There used to be a similar problem when the feature was implemented with generate_systemd options in container https://github.com/containers/ansible-podman-collections/issues/319 but it was fixed. However, the new module does not seem to support that when I tested it.

**Steps to reproduce the issue:**

1. Run some playbook like this:
```
- hosts: server
  tasks:
    - name: HTTP Echo Container
        name: echo
        image: mendhak/http-https-echo
        state: "started"
    - name: Setup systemd
      containers.podman.podman_generate_systemd:
        container_prefix: ""
        dest: "{{ ansible_user_dir }}/.config/systemd/user"
        name: echo
        pod_prefix: ""
        separator: ""
      register: output
    - debug:
        var: output
```

**Describe the results you received:**

- The service file is created, but with default parameter name. For example, the above will create **container-echo.service**.
- The output of the `podman_generate_systemd` task is:
```
TASK [debug] *********************************************************************************************************************************************************************************************************************************************************************************************************************************
ok: [server.lan] => {
    "output": {
        "changed": true,
        "failed": false,
        "podman_command": "podman generate systemd --name --format json echo",
        "systemd_units": {
            "container-echo": "# container-echo.service\n# autogenerated by Podman 4.3.1\n# Mon Nov 21 21:08:17 +07 2022\n\n[Unit]\nDescription=Podman container-echo.service\nDocumentation=man:podman-generate-systemd(1)\nWants=network-online.target\nAfter=network-online.target\nRequiresMountsFor=/run/user/1000/containers\n\n[Service]\nEnvironment=PODMAN_SYSTEMD_UNIT=%n\nRestart=on-failure\nTimeoutStopSec=70\nExecStart=/usr/bin/podman start echo\nExecStop=/usr/bin/podman stop  \\\n\t-t 10 echo\nExecStopPost=/usr/bin/podman stop  \\\n\t-t 10 echo\nPIDFile=/run/user/1000/containers/overlay-containers/e3d79ebd42b66eae4b5a81052a4f177dcde46af741a365065c35e23855a3f1f8/userdata/conmon.pid\nType=forking\n\n[Install]\nWantedBy=default.target\n"
        }
    }
}
```
- From the output, the podman command `podman generate systemd --name --format json echo` seems to not have any of the `prefix` options when the string is empty

**Describe the results you expected:**

The example above, with the podman generate systemd command and empty string

```
podman generate systemd --name --files --container-prefix "" --pod-prefix "" --separator "" echo
```
The generated file will be named **echo.service**

zephyros-dev:
Fixes #516

### DIFF
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -273,7 +273,7 @@ def generate_systemd(module):
 
     #  Container-prefix option
     container_prefix = module.params['container_prefix']
-    if container_prefix:
+    if container_prefix is not None:
         command_options.append(
             '--container-prefix={container_prefix}'.format(
                 container_prefix=container_prefix,
@@ -282,7 +282,7 @@ def generate_systemd(module):
 
     #  Pod-prefix option
     pod_prefix = module.params['pod_prefix']
-    if pod_prefix:
+    if pod_prefix is not None:
         command_options.append(
             '--pod-prefix={pod_prefix}'.format(
                 pod_prefix=pod_prefix,
@@ -291,7 +291,7 @@ def generate_systemd(module):
 
     #  Separator option
     separator = module.params['separator']
-    if separator:
+    if separator is not None:
         command_options.append(
             '--separator={separator}'.format(
                 separator=separator,


sshnaidm:
Thanks @zephyros-dev !

Issue #529, opened 2022-11-29
Podman unshare `Error: please use unshare with rootless`


Igortorrente:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**

Following the instruction in the [documentation](https://github.com/containers/ansible-podman-collections/pull/455/files#diff-628ad9cab176f2b22c27386d21d0acc2a368fc9836ffb67a2a63828bddb7a7ffR95-R106) I tried the new podman unshare `become_method`, but it didn't worked for me.

It seems to be running `sudo podman unshare <cmd>` under the hood instead of `podman unshare <cmd>` for some reason.

**Steps to reproduce the issue:**

1. Create a task with podman unshare

2. Run it

3. Face the error message

**Describe the results you received:**

``` yaml
failed: [instance] (item=/home/vagrant/containers_cache/redis) => {"ansible_loop_var": "item", "changed": false, "item": "/home/vagrant/containers_cache/redis", "module_stderr": "", "module_stdout": "Error: please use unshare with rootless\r\n", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", "rc": 125}
failed: [instance] (item=/home/vagrant/containers_config/redis/) => {"ansible_loop_var": "item", "changed": false, "item": "/home/vagrant/containers_config/redis/", "module_stderr": "", "module_stdout": "Error: please use unshare with rootless\r\n", "msg": "MODULE FAILURE\nSee stdout/stderr for the exact error", "rc": 125}
```

**Describe the results you expected:**

Success

**Or version from `ansible-galaxy` if installed from galaxy: `ansible-galaxy collection list | grep containers.podman`**

``` Console
$ ansible-galaxy collection list | grep containers.podman
containers.podman             1.9.4  
```

**Output of `ansible --version`:**

``` Console
$ ansible --version
ansible [core 2.13.5]
  config file = /home/igor/projects/cloud/Caedrium-Playbooks/ansible.cfg
  configured module search path = ['/home/igor/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /home/igor/projects/cloud/python3_venv/lib/python3.10/site-packages/ansible
  ansible collection location = /home/igor/.ansible/collections:/usr/share/ansible/collections
  executable location = /home/igor/projects/cloud/python3_venv/bin/ansible
[...]

PR #530, opened 2022-11-30
[FIX] Issue #529: become plugin podman_unshare become_user default

grzs:
Signed-off-by: Janos Gerzson <gerzsonj@gmail.com>

### DIFF
--- a/plugins/become/podman_unshare.py
+++ b/plugins/become/podman_unshare.py
@@ -17,8 +17,7 @@ DOCUMENTATION = """
     version_added: 1.9.0
     options:
         become_user:
-            description: User you 'become' to execute the task
-            default: root
+            description: User you 'become' to execute the task ('root' is not a valid value here).
             ini:
               - section: privilege_escalation
                 key: become_user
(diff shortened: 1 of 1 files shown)


plugins/become/podman_unshare.py
```diff
@@ -130,8 +129,8 @@ def build_become_command(self, cmd, shell):
 
         becomecmd = 'podman unshare'
 
-        user = self.get_option('become_user') or ''
-        if user:
+        user = self.get_option('become_user') or 'root'
```
sshnaidm, line 132: Shouldn't it be the current user by default?

grzs: It will run as the current user (ansible_user) if become_user is root, or not set at all (by default).
Actually, if become_user has been set to anything but root it will be executed with sudo, otherwise no sudo applied.
I tested it and was working flawlessly.

plugins/become/podman_unshare.py
```diff
@@ -130,8 +129,8 @@ def build_become_command(self, cmd, shell):
...
+        if user != 'root':
```
grzs, line 133: here is the modified test

plugins/become/podman_unshare.py
sshnaidm: Thanks for clarification. I think we can add some tests here.
I'll prepare a test skeleton.

grzs: ok

sshnaidm: You can use this as a template for tests: https://github.com/containers/ansible-podman-collections/pull/531 Just replace `podman_lovely_module` by your module/plugin name everywhere.

PR #533, opened 2022-12-12
Fix undesirable splitting of IPv6 host addresses

linxside:
Currently, the podman_container module allows setting explicit `etc_hosts`. A host can also have an IPv6 address. When a host contains any IPv6 address, the module creates a container with exactly that config.
```yaml
- name: Bug example container
    image: docker.io/library/debian
    name: test
    etc_hosts:
      "example.org": "fd00::1"
    state: present
```
If we run the play again, the podman_container module will break with the following error:
`ValueError: dictionary update sequence element #0 has length 4; 2 is required` in `diffparam_etc_hosts`.

What happens is the data returned from podman contains multiple colons, due to the IPv6 address. The current code splits the returned string between every colon, which is a bit too much. My fix for it is to restrict the splitting only for the first colon from the left hand, so the hostname and IP parts are split correctly.


sshnaidm:
Thanks for patch and for test in particular!

Issue #352, opened 2021-12-06
Provide podman_prune module


Nothing4You:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind feature

**Description**

This was already briefly mentioned in https://github.com/containers/ansible-podman-collections/discussions/307#discussioncomment-1375590, however, there isn't an official feature request for this until now.

Similar to `docker_prune`, this would allow removing various items, such as unused images.
Currently if you regularly deploy new images as far as I can tell there is no way to clean up old images other than running podman CLI e.g. via `command` module rather than using a module from this collection.

PR #538, opened 2023-01-16
added podman_prune module

rebtoor:
Added podman_prune module, it should fix https://github.com/containers/ansible-podman-collections/issues/352

WIP since i need to add some tests

### DIFF
--- a/plugins/modules/podman_prune.py
+++ b/plugins/modules/podman_prune.py
@@ -15,7 +15,8 @@ version_added: '1.10.0'
 short_description: Allows to prune various podman objects
 notes: []
 description:
-    - Allows to run C(podman container prune), C(podman image prune), C(podman network prune), C(podman volume prune) and C(podman system prune)
+    - Allows to run C(podman container prune), C(podman image prune), C(podman network prune), 
+      C(podman volume prune) and C(podman system prune)
 requirements:
     - "Podman installed on host"
 options:
@@ -33,7 +34,8 @@ options:
         description:
             - A dictionary of filter values used for selecting containers to delete.
             - "For example, C(until: 24h)."
-            - See L(the podman documentation, https://docs.podman.io/en/latest/markdown/podman-container-prune.1.html#filter-filters)
+            - See L(the podman documentation, 
+              https://docs.podman.io/en/latest/markdown/podman-container-prune.1.html#filter-filters)
               for more information on possible filters.
         type: dict
     image:
@@ -45,7 +47,8 @@ options:
         description:
             - A dictionary of filter values used for selecting images to delete.
             - "For example, C(dangling: true)."
-            - See L(the podman documentation,https://docs.podman.io/en/latest/markdown/podman-image-prune.1.html#filter-filters)
+            - See L(the podman documentation,
+              https://docs.podman.io/en/latest/markdown/podman-image-prune.1.html#filter-filters)
               for more information on possible filters.
         type: dict
     network:
@@ -56,7 +59,8 @@ options:
     network_filters:
         description:
             - A dictionary of filter values used for selecting networks to delete.
-            - See L(the podman documentation,https://docs.podman.io/en/latest/markdown/podman-network-prune.1.html#filter)
+            - See L(the podman documentation,
+              https://docs.podman.io/en/latest/markdown/podman-network-prune.1.html#filter)
               for more information on possible filters.
         type: dict
     system:
@@ -82,7 +86,8 @@ options:
     volume_filters:
         description:
             - A dictionary of filter values used for selecting volumes to delete.
-            - See L(the podman documentation,https://docs.podman.io/en/latest/markdown/podman-volume-prune.1.html#filter)
+            - See L(the podman documentation,
+              https://docs.podman.io/en/latest/markdown/podman-volume-prune.1.html#filter)
               for more information on possible filters.
         type: dict
 """
@@ -177,7 +182,7 @@ def podmanExec(module, target, filters, executable):
     return changed, out, err
 
 
-def run_module():
+def main():
     module_args = dict(
         container=dict(type='bool', default=False),
         container_filters=dict(type='dict'),
@@ -283,10 +288,6 @@ def run_module():
         )
 
     module.exit_json(**results)
-
-
-def main():
-    run_module()
 
 
 if __name__ == '__main__':
--- a/plugins/modules/podman_prune.py
+++ b/plugins/modules/podman_prune.py
@@ -1,0 +1,293 @@
+#!/usr/bin/python
+# -*- coding: utf-8 -*-
+
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+# Copyright (c) 2023, Roberto Alfieri <ralfieri@redhat.com>
+
+from __future__ import absolute_import, division, print_function
+__metaclass__ = type
+
+DOCUMENTATION = r"""
+module: podman_prune
+author:
+    - "Roberto Alfieri (@rebtoor)"
+version_added: '1.10.0'
+short_description: Allows to prune various podman objects
+notes: []
+description:
+    - Allows to run C(podman container prune), C(podman image prune), C(podman network prune), C(podman volume prune) and C(podman system prune)
+requirements:
+    - "Podman installed on host"
+options:
+    executable:
+        description:
+            - Podman binary.
+        type: str
+        default: podman
+    container:
+        description:
+            - Whether to prune containers.
+        type: bool
+        default: false
+    container_filters:
+        description:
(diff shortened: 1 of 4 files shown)


sshnaidm:
Please use this skeleton for tests: https://github.com/containers/ansible-podman-collections/pull/531

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,293 @@
...
+        changed, out, err = podmanExec(module, target, filters, executable)
+        if not out:
+            containers = []
+        else:
+            containers = out.rstrip().split("\n")
+        results = dict(
+            changed=changed,
+            containers=containers,
+            stderr=err
+        )
```
sshnaidm, line 218: I see this pattern repeats every time, can we maybe to optimize a little bit? Like:
```python
for target, filters in (("container", "container_filters"), ("network", "network_filters"), ("image", "image_filters"), ("volume", ("volume_filters")):
    if module.params[target]:
        changed, out, err = podmanExec(module, target, module.params[filters], executable)

File: plugins/modules/podman_prune.py [hunk 2]
```diff
@@ -0,0 +1,293 @@
...
+    changed = False
+    command = [executable, target, 'prune', '--force']
+    if filters != "" and target != "system":
+        command.append("--filter=")
+        command.append(filters)
+    if filters != "" and target == "system":
+        split_filters = filters.strip().split(" ")
+        for filter in split_filters:
+            if filter:
+                command.append(filter)
```
sshnaidm, line 170: filters is better to process in main if it's only for "system"

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,293 @@
...
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def podmanExec(module, target, filters, executable):
+    changed = False
+    command = [executable, target, 'prune', '--force']
+    if filters != "" and target != "system":
+        command.append("--filter=")
+        command.append(filters)
```
sshnaidm, line 165: probably need to convert filters to "key-value", because it's `dict` in module params

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,293 @@
...
+    if filters != "" and target == "system":
+        split_filters = filters.strip().split(" ")
```
sshnaidm, line 167: filters is a dictionary, not a string

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,293 @@
...
+        for filter in split_filters:
```
sshnaidm, line 168: looping over filters should be for all options, not only for system, also containers can get multiple filters

sshnaidm (changes requested):
Starting the review, a few comments inside.

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,293 @@
...
+    rc, out, err = module.run_command(command)
+    if out:
+        changed = True
+    if rc != 0:
+        module.fail_json(
+            msg="Error executing prune on {target}: {err}".format(target=target, err=err))
+    return changed, out, err
+
+
+def run_module():
```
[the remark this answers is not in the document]
rebtoor: I took inspiration from [Developing modules](https://docs.ansible.com/ansible/latest/dev_guide/developing_modules_general.html) doc, but you're right.

rebtoor:
Hi @sshnaidm , I followed your advices on PR and tried to refactor some things inside code. I enabled filters parsing for *all* modules (in both forms, e.g. "--filter=until=2h" and "--filter=label=Maintainers=Podman Maintainers") and handled the special case of "image" module that can use 2 more flags: "dangling_only" (-a flag in original command) and "external" (--external flag in original command). I'll wait for your review and try to implement some tests.

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,247 @@
...
+                                                                         value=filters[common_filter]))
+        else:
+            for system_filter in filters:
+                command.append(filters[system_filter])
+
+    rc, out, err = module.run_command(command)
+    if out:
+        changed = True
+    else:
+        changed = False
```
sshnaidm, line 193:
```suggestion
    changed = bool(out)
```

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,247 @@
...
+  sample: []
+'''
+
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def podmanExec(module, target, filters, executable):
+    command = [executable, target, 'prune', '--force']
+    if filters:
```
sshnaidm, line 168:
```suggestion
    filters = filters or {}
```
We can eliminate one level of `if` nesting. Looping over empty dictionary will be just skipped.

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,247 @@
...
+                        if common_filter == 'dangling_only' and not filters['dangling_only']:
+                            command.append('-a')
+                        if common_filter == 'external' and filters['external']:
+                            command.append('--external')
+                    else:
+                        command.append('--filter={label}={value}'.format(label=common_filter,
+                                                                         value=filters[common_filter]))
+        else:
+            for system_filter in filters:
+                command.append(filters[system_filter])
```
sshnaidm, line 187: For better readability it's better to put it above with `if target == 'system':` and all the rest with `else`

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,247 @@
...
+  type: list
+  elements: str
+  sample: []
+'''
+
+
+from ansible.module_utils.basic import AnsibleModule
+
+
+def podmanExec(module, target, filters, executable):
```
sshnaidm, line 166: I'd recommend to to filters preparing in a separate function, and to leave here only podman execution.
Then you can jump earlier from filters function and avoid deep nesting.
Also filters processing is completely independent, so it's right to do it separately.

plugins/modules/podman_prune.py
```diff
@@ -0,0 +1,247 @@
...
+        changed = False
+    if rc != 0:
+        module.fail_json(
+            msg='Error executing prune on {target}: {err}'.format(target=target, err=err))
+
+    return {
+        "changed": changed,
+        "stdout": out,
+        "stderr": err
+    }
```
sshnaidm, line 202: Actually this part should be in podman_exec function, filter prepare in a separate.

sshnaidm (changes requested):
Thanks, looks much better! A few another comment inside, worth to split the function to two.

sshnaidm:
Great, thanks!
Let's create tests now.

rebtoor:
> Great, thanks! Let's create tests now.

done! I've also handled output when multiple modules are invoked all at once, it came out during tests... :D

tests/integration/targets/podman_prune/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+        network: true
+        volume: true
+      register: result
+
+    # Analyze result
+    - name: Show results
+      ansible.builtin.debug:
+        var: result
+
+    - name: Verify assertions for network, container and volume
```
sshnaidm, line 40: I think would be useful to check that container, volume and network don't exist after pruning. Because in CI nobody is looking on debug print :) We can check this either by shell "podman exists volume/network/container or by info modules: `containers.podman.podman_container_info`, `containers.podman.podman_network_info`, `containers.podman.podman_volume_info` and see they have empty results.

tests/integration/targets/podman_prune/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+    # Test with filters
+    - name: Prune objects with filters
+      containers.podman.podman_prune:
+        image: true
+        image_filters:
+          dangling_only: false
+          external: true
+      register: image_filters
+
+    - name: Verify assertions for image (with filters)
```
sshnaidm, line 59: and here as well.

tests/integration/targets/podman_prune/tasks/main.yml
```diff
@@ -0,0 +1,100 @@
...
+        system: true
+        system_all: true
+        system_volumes: true
+      register: system_result
+
+    - name: Show results
+      ansible.builtin.debug:
+        var: system_result
+
+    - name: Verify assertions for system
```
sshnaidm, line 92: and here

sshnaidm (changes requested):
Thanks! A few comments about tests.
Please rebase your patch on latest.

tests/integration/targets/podman_prune/tasks/main.yml
rebtoor: ack!

sshnaidm (approved):

sshnaidm:
Thanks for your contribution!

PR #542, opened 2023-01-21
Improved documentation of `podman_generate_systemd` module

exploide:
Improved documentation of `podman_generate_systemd` module.

The first commit fixes some typos. Most notably the command syntax is `systemctl daemon-reload` with a `-` in between. Furthermore I moved the hints regarding typical systemd unit folders to the description of the `dest` parameter, because that's where this info is needed.

The second commit changed an example because it took me some time to find out why my playbook crashed on the second run. It looked much like #345 and it turns out, while the `generate_systemd` part of the `container` module automatically adds `rm: true` when `new: true` is given, this behavior seems to be impossible for the new `podman_generate_systemd` module. On container creation it cannot know that the `podman_generate_systemd` module is used with `new: true` afterwards.
So I wanted to document that we need to manually specify `rm: true`. I thought about adding another example or changing the existing one. I opted for the latter to keep the documentation short and comprehensive.

Please tell me if I should change that. But actually I don't even understand why the module is used by some people without `new: true` and `no_header: true` at all. Isn't it basically impossible to create an idempotent configuration as code definition of your container services without doing it this way? I don't really understand the use case I guess.

Besides, I was not sure if I need to regenerate the html docs right now or if this will happen later. Just tell me if this is required.

### DIFF
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -137,6 +137,31 @@ notes:
 '''
 
 EXAMPLES = '''
+# Example of creating a container and systemd unit file.
+# When using podman_generate_systemd with new:true then
+# the container needs rm:true for idempotence.
+- name: Create postgres container
+  containers.podman.podman_container:
+    name: postgres
+    image: docker.io/library/postgres:latest
+    rm: true
+    state: created
+
+- name: Generate systemd unit file for postgres container
+  containers.podman.podman_generate_systemd:
+    name: postgres
+    new: true
+    no_header: true
+    dest: /etc/systemd/system
+
+- name: Ensure postgres container is started and enabled
+  ansible.builtin.systemd:
+    name: container-postgres
+    daemon_reload: yes
+    state: started
+    enabled: yes
+
+
 # Example of creating a container and integrate it into systemd
 - name: A postgres container must exist, stopped
   containers.podman.podman_container:
--- a/plugins/modules/podman_generate_systemd.py
+++ b/plugins/modules/podman_generate_systemd.py
@@ -23,7 +23,9 @@ options:
     required: true
   dest:
     description:
-      - Destination of the generated systemd unit file(s)
+      - Destination of the generated systemd unit file(s).
+      - Use C(/etc/systemd/system) for the system-wide systemd instance.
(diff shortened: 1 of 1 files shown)


sshnaidm:
@CyberFox001 FYI

CyberFox001:
The first commit is nice, but I have difficulty to understand the second one.

If you create a container who is started, then generate the systemd unit file and finally ask systemd to start your new service: You will get an error. Because systemd don't know that the container is already started, it will try do start it and get an error from Podman.

I think the necessity to set `rm: true` on `podman_container` when you want to set `new: true` for `podman_generate_systemd` is another use case. My example was about creating the container with the Ansible playbook (or role) and manage it start/stop/restart with Systemd. But a hint and another example about the necessity of `rm` option on `podman_container` when you want to set `new` option for `podman_generate_systemd` is welcome. :)

exploide:
> If you create a container who is started, then generate the systemd unit file and finally ask systemd to start your new service: You will get an error. Because systemd don't know that the container is already started, it will try do start it and get an error from Podman.

What I create there is a container in state `created` not `started`. Starting the container is done by systemd afterwards. Where would you expect the error?

> My example was about creating the container with the Ansible playbook (or role) and manage it start/stop/restart with Systemd.

That's also my intention. But I'm using `new: true` here to have the unit files being less strict on container IDs and the like. The reason is that I plan to use `podman auto-update` and therefore the unit files need to be flexible enough.

I you prefer that I restore the previous example and add a new one, that would be totally fine. But I would like to clarify first if we have a misunderstanding or where you see the difference.

In the end, I'm also interested what advantage the unit generation without `new: true` has. I only see disadvantages so far.

exploide:
I amended my commit, restored the previous example and added my idempotent one as a separate example. I guess this is fine in any case.

sshnaidm:
@exploide can you please fix CI sanity job?
```
ERROR: Found 1 pep8 issue(s) which need to be resolved:
ERROR: plugins/modules/podman_generate_systemd.py:27:161: E501: line too long (184 > 160 characters)
```

exploide:
Done. I think it's fixed now.

sshnaidm:
@exploide thanks for your contribution!

PR #472, opened 2022-09-02
Support passing multiple networks with params

sshnaidm:
Fix #467
Signed-off-by: Sagi Shnaidman <sshnaidm@redhat.com>

PR #553, opened 2023-02-23
pod: Support passing multiple networks with params

BenjaminSchubert:
Similar to #472, but for pods. there is also support there

I wasn't sure if/how to add tests there, I'd be happy to if I can get some guidance


sshnaidm (approved):
Great, thanks for contribution!

Issue #565, opened 2023-03-30
podman_login - cannot use module without registry defined


spmfox:
There is an issue where the 'registry' parameter is effectively required to use the podman_login module. Current manual says the 'registry' parameter is optional, and I believe it should continue to be optional.

How to reproduce:
1. Use module with no 'registry' defined
2. Login succeeds first time
3. Run module again
4. Fails the second time with ```TypeError: encoding without a string argument```

The issue appears to be here:
``` python
121    if os.path.exists(check_file):
122        content = open(check_file, 'rb').read()
123        if bytes(registry, 'UTF-8') in content:
124            checksum = hashlib.md5(content).hexdigest()
125    rc, out, err = module.run_command(command)

136        if checksum:
137            content = open(check_file, 'rb').read()
138            if bytes(registry, 'UTF-8') in content:
139                new_checksum = hashlib.md5(content).hexdigest()
```

If we are checking the contents of the authfile against the defined registry, and this breaks if no registry is defined.

I believe we can just remove the check for the registry text inside of the authfile. The outcome is still the same, the authfile is hashed before and after the ```podman login``` is issued. There seems to be nothing gained by checking for the registry inside of the authfile anyway, especially since the registry can be defined elsewhere outside of this module.

spmfox:
I opened a pull request to address this issue:
https://github.com/containers/ansible-podman-collections/pull/566

sshnaidm:
Closed by https://github.com/containers/ansible-podman-collections/pull/566

Issue #564, opened 2023-03-30
podman_login - does not work in a FIPS environment


spmfox:
The following code causes the podman_login module to not work at all in FIPS compliant environments:

``` python
124            checksum = hashlib.md5(content).hexdigest()

139                new_checksum = hashlib.md5(content).hexdigest()
```

error:
```
line 123, in login
ValueError: [digital envelope routines: EVP_DigestInit_ex] disabled for FIPS
```

This can be fixed by changing the hash to something that is allowed in FIPS environments, such as sha256.

spmfox:
I opened a pull request to address this issue:
https://github.com/containers/ansible-podman-collections/pull/566

sshnaidm:
Closed by https://github.com/containers/ansible-podman-collections/pull/566

PR #566, opened 2023-03-30
podman-login - fix FIPS md5 issue and registry requirement

spmfox:
This addresses two issues:
https://github.com/containers/ansible-podman-collections/issues/564
https://github.com/containers/ansible-podman-collections/issues/565


sshnaidm:
Great, thanks for contribution!

PR #582, opened 2023-04-21
fix(podman-play): fix regex for pod kube recreate

kuhball:
This PR fixes the regex used when no `yaml` import is successful. When using the `recreate: true` option without `yaml` the error `Deployment doesn't have a name!` occured.


sshnaidm:
Thanks for catching this!

PR #599, opened 2023-06-11
Fix idempotency checks when using `userns`

BenjaminSchubert:
Previously, this would only mark the userns as expected when `uidmap` or `gidmap` is passed. However, in the case `userns` is passed, this is also expected to be set.

### DIFF
--- a/tests/integration/targets/podman_pod/tasks/main.yml
+++ b/tests/integration/targets/podman_pod/tasks/main.yml
@@ -333,7 +333,7 @@
         that:
           - pod15_a_info is changed
 
-    - name: Start pod with share for idempotency
+    - name: Start pod with gidmap and uidmap for idempotency
       containers.podman.podman_pod:
         executable: "{{ test_executable | default('podman') }}"
         name: pod1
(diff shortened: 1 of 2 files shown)


tests/integration/targets/podman_pod/tasks/main.yml
```diff
@@ -347,6 +347,47 @@
         that:
           - pod15_b_info is not changed
 
+    # Reset the namespace check.
```
BenjaminSchubert, line 350: Reviewers: I am unsure of what to do here. This looks like another bug with regards to idempotency, it doesn't seem `podman inspect` has any info that would allow figuring out which is which.

BenjaminSchubert:
I am also unsure what is happening with the tripleo ci here, it doesn't seem related to my changes? (Or I don't know how to read that output. If someone has clues, that would be appreciated

tests/integration/targets/podman_pod/tasks/main.yml
sshnaidm: Probably we should use `CreateCommand` to understand if `userns` was passed to Podman.

```json
"CreateCommand": [
            "podman",
            "pod",
            "create",
            "--name",
            "pod2",
            "--userns",
            "auto"
        ],
```
Like we do here for example: https://github.com/containers/ansible-podman-collections/blob/0ae6baac3d673da1d1d545f2e971152fc2b6a8d3/plugins/module_utils/podman/podman_container_lib.py#L1068C1-L1071
Maybe worth to open issue to Podman itself, looks like a bug in inspection for me.
Anyway, this change is OK.

sshnaidm (approved):

sshnaidm:
@BenjaminSchubert can you please rebase on latest master?
Thanks for the fix!

sshnaidm:
> I am also unsure what is happening with the tripleo ci here, it doesn't seem related to my changes? (Or I don't know how to read that output. If someone has clues, that would be appreciated

You can ignore it, I work on removing them.

BenjaminSchubert:
@sshnaidm now rebased!

tests/integration/targets/podman_pod/tasks/main.yml
sshnaidm: Yeah, I think it's only an option in such cases.
Of course it doesn't prevent you to submit issues upstream anyway :)

Issue #609, opened 2023-07-30
Podman volume export module


DilasserT:
/kind feature

**Description**

Hi, 

Would it be possible to export volumes with a module similar to `podman_export` ?
Or maybe it should be an additional option to `podman_volume`.

I did not see an option to do this specifically:
`podman volume export myvolume --output myvolume.tar`

I was expecting something like:

```
- containers.podman.podman_volume_export:
    dest: /path/to/myvolume.tar
    volume: myvolume
```

sshnaidm:
I think adding to current `podman_export` would be easier. Want to do a patch?

DilasserT:
I can try. How would you implement it ?

Maybe adding a `volume` field instead of the `container` field, with both fields exclusive ?

```
- containers.podman.podman_export:
    dest: /path/to/myvolume.tar
    volume: myvolume
```

sshnaidm:
I think we can add *in addition* not instead. Both exclusive is fine with me.

DilasserT:
Would it make sense to add the same mechanic to `podman_load` in the same PR ?

sshnaidm:
> Would it make sense to add the same mechanic to `podman_load` in the same PR ?

Yeah, totally.

sshnaidm:
Done in https://github.com/containers/ansible-podman-collections/pull/617

PR #617, opened 2023-08-07
Adding volume import and export option

DilasserT:
adding volume import and volume export to podman_import and podman_export 
Updating integration tests
Issue #609

### DIFF
--- a/plugins/modules/podman_import.py
+++ b/plugins/modules/podman_import.py
@@ -29,6 +29,10 @@ options:
     - Set changes as list of key-value pairs, see example.
     type: list
     elements: dict
+  volume:
+    description:
+    - Volume to import, cannot be used with change and commit_message
+    type: str
   executable:
     description:
       - Path to C(podman) executable if it is not in the C($PATH) on the
@@ -95,8 +99,12 @@ EXAMPLES = '''
       - "CMD": /bin/bash
       - "User": root
     commit_message: "Importing image"
+- containers.podman.podman_import:
+    src: /path/to/tar/file
+    volume: myvolume
 '''
 
+import os  # noqa: E402
 import json  # noqa: E402
 from ansible.module_utils.basic import AnsibleModule  # noqa: E402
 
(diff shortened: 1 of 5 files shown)


plugins/modules/podman_import.py
```diff
@@ -127,29 +135,65 @@ def load(module, executable):
         module.fail_json(msg="Could not parse JSON from image %s: %s" % (image_name, e))
     return changed, out, err, info, command
 
+def volume_load(module, executable):
+    changed = False
+    command = [executable, 'volume', 'import', module.params['volume'], module.params['src']]
+    src = module.params['src']
+    if os.path.exists(src):
```
sshnaidm, line 142: We can let Podman handle this, in line 147 below it will fail when not finding the requested volume, with appropriate message.

plugins/modules/podman_import.py
```diff
@@ -127,29 +135,65 @@ def load(module, executable):
...
+        changed = True
```
sshnaidm, line 143: In this case it's always `True`, or module fails, I don't see an option it exits with `changed=False`. Probably it should check if current volume exists and not to do anything if it exists? I'm not sure, because Podman imports volume in any case and overrides the existing ones.

plugins/modules/podman_import.py
```diff
@@ -127,29 +135,65 @@ def load(module, executable):
...
+        changed, out, err, image_info, command = load(module, executable)
+        import_object = 'image'
+        import_info = image_info
 
     results = {
         "changed": changed,
         "stdout": out,
         "stderr": err,
-        "image": image_info,
+        import_object: import_info,
```
sshnaidm, line 192: Let's keep `image` for backward compatibility, people might using it in their playbooks. I think we can just add `"volume": volume_info` in addition?

sshnaidm (changes requested):
Thanks! A few comments on the patch.

sshnaidm:
And also logs from CI, please check:
```
ERROR: plugins/modules/podman_export.py:94:13: E128: continuation line under-indented for visual indent
ERROR: plugins/modules/podman_export.py:97:1: E302: expected 2 blank lines, found 1
ERROR: plugins/modules/podman_export.py:108:10: E121: continuation line under-indented for hanging indent
ERROR: plugins/modules/podman_export.py:111:9: E122: continuation line missing indentation or outdented
ERROR: plugins/modules/podman_export.py:114:1: W293: blank line contains whitespace
ERROR: plugins/modules/podman_import.py:138:1: E302: expected 2 blank lines, found 1
ERROR: plugins/modules/podman_import.py:160:1: E302: expected 2 blank lines, found 1
ERROR: plugins/modules/podman_import.py:170:10: E121: continuation line under-indented for hanging indent
ERROR: plugins/modules/podman_import.py:197:5: E303: too many blank lines (2)
ERROR: plugins/modules/podman_export.py:114:0: trailing-whitespace: Trailing whitespace
ERROR: plugins/modules/podman_import.py:151:70: undefined-variable: Undefined variable 'image_name'
ERROR: plugins/modules/podman_import.py:155:78: undefined-variable: Undefined variable 'image_name'
ERROR: plugins/modules/podman_export.py:0:0: doc-required-mismatch: Argument 'container' in argument_spec is not required, but is documented as being required
```

plugins/modules/podman_import.py
DilasserT: So either it is always true, or we need to check somehow that the existing volume is the same as the tarball ?
Is there a way to do that ?

sshnaidm: I think it's an overkill to check it. Podman just overrides, probably we need to follow this logic (whatever it's right or not), so we'll have it always changed=true.

sshnaidm (approved):

sshnaidm:
Awesome! Can you just add signature to commit to make DCO job happy? It can be done with `git commit --amend -s`.
Thanks

DilasserT:
@sshnaidm Just commenting in case you did not see the update.

sshnaidm:
Awesome, merging. Thanks for contribution!

PR #624, opened 2023-08-13
Add a force field to podman_generate_systemd

### LINKED ISSUE #606: Add a `force` field to `podman_generate_systemd`

/kind feature

**Description**

Currently `podman_generate_systemd` creates a systemd service unit file only if it doesn't exist yet on a given path.
It would be useful if we could specify a parameter to replace the existing file with a new one.
The name of the field will be `force` like it in `ansible.builtin.copy` module, and `false` will be the default for backward compatibility. What do you think of it?

Ex.
```
   # Replace systemd unit file if it exist on the same path
    - name: Systemd unit files for postgres container must exist
      containers.podman.podman_generate_systemd:
        name: postgres_local
        dest: ~/.config/systemd/user/
        force: true
```

nishipy:
Fixes #606


sshnaidm:
Thanks!

PR #630, opened 2023-08-22
Add CI for runlabel testing

sshnaidm:


pdostal:
I have only one question: Why there is so much stuff defined in the `Dockerfile` and then never used? The container is just built, not even executed.

sshnaidm:
@pdostal I just copied it from a different test :shrug: https://github.com/containers/ansible-podman-collections/blob/master/tests/integration/targets/podman_container_idempotency/files/Dockerfile

PR #632, opened 2023-08-22
Add target support for podman build image

### LINKED ISSUE #631: Add support for targets in multi-stage builds
Requesting support for targets (--target) to be added to the image module for builds (https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_image_module.html).  

This is currently not a feature in the podman module and is necessary for building multi-stage images.

Example:
```
- name: Build image {{ build_image }} with podman
  containers.podman.podman_image:
    name: "{{ build_image }}"
    path: src/{{ clone_repo }}/{{ clone_branch }}
    build:
      target: eda-ui
  register: podman_build
  until: podman_build is not failed
  retries: 1
  delay: 10
```
  

That would do the same as a `podman build --target eda-ui` (using https://github.com/ansible/ansible-ui/blob/main/Dockerfile) as an example Containerfile.

sshnaidm:
Fix #631

### DIFF
--- a/plugins/modules/podman_image.py
+++ b/plugins/modules/podman_image.py
@@ -132,6 +132,10 @@ DOCUMENTATION = r'''
           description:
             - Extra args to pass to build, if executed. Does not idempotently check for new build args.
           type: str
+        target:
+          description:
+            - Specify the target build stage to build.
+          type: str
     push_args:
       description: Arguments that control pushing images.
       type: dict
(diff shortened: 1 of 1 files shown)


aknochow:
Looks good:

play:
```- name: Build image with podman and target
  hosts: localhost
  vars:
    build_image: test:podman
  tasks:

  - name: Build image {{ build_image }} with podman
    containers.podman.podman_image:
      name: "{{ build_image }}"
      path: /stage
      build:
        target: eda-ui
    register: podman_build
    until: podman_build is not failed
    retries: 1
    delay: 10
```

resulting command:
```
    "podman_actions": [
        "/usr/bin/podman image ls test:podman --format json",
        "/usr/bin/podman image exists test:podman",
        "/usr/bin/podman build -t test:podman --format oci --rm --target eda-ui /stage",
        "/usr/bin/podman inspect c7d17ccbc6e4 --format json"
    ],

PR #669, opened 2023-11-21
Recreate stopped containers if `recreate` flag is enabled

rebtoor:
If `recreate` flag is enabled we should be able to recreate stopped containers even if their configuration aren't changed.

This become important if those containers has bind mounted files that are recreated from scratch instead of being edited (for example files managed by ansible or puppet)


rebtoor:
> Thanks for contribution! Can we add a test for such a case in our test cases? https://github.com/containers/ansible-podman-collections/blob/master/tests/integration/targets/podman_container/tasks/main.yml

Ack! I'll send it ASAP

PR #673, opened 2023-11-27
Improve `security_opt` comparison between existing containers

rebtoor:
Since SElinux labels are basically annotations[1], they are merged in a single comma separated string in the list by podman, so we need to split them in a sorted list if we want to compare it to the list that we provide to the module.

Also, a proper test for this case has been added.

[1] https://github.com/containers/podman/blob/d1236f46fc5cb29e8da296c3662f8ce869865030/libpod/define/annotations.go#L49-L54


sshnaidm:
Great! Thanks for contribution!

PR #677, opened 2023-11-29
Add support setting underlying interface in podman_network

pablomh:
It may be desirable to set a different interface than the default one (for example, a different bridge).

In the case of a macvlan interface, it can be used instead of setting `opt.parent`.


sshnaidm:
Thanks for contribution!

PR #687, opened 2023-12-11
Fix podman_login.py example name

Apollo3zehn:
The second example in `podman_login.py` show the login for quay.io instead of the default registry.


sshnaidm:
Thanks, please sign the commit with `git -s` to make DCO job happy.

Apollo3zehn:
I hope it is fine now, had to figure out how to use my SSH key for signing (and I used `git -S` before realizing it should be `git -s`).

sshnaidm:
Cool, it's good now.

PR #694, opened 2023-12-22
containers.podman.podman_container: Add env_files param

ant-bl:
Add a env_files parameter to use several file to set environment variables.

### DIFF
--- a/plugins/module_utils/podman/common.py
+++ b/plugins/module_utils/podman/common.py
@@ -306,3 +306,15 @@ def normalize_signal(signal_name_or_number):
         if signal_name not in _signal_map:
             raise RuntimeError("Unknown signal '{0}'".format(signal_name_or_number))
         return str(_signal_map[signal_name])
+
+
+def get_podman_version(module, fail=True):
+    executable = module.params['executable'] if module.params['executable'] else 'podman'
+    rc, out, err = module.run_command(
+        [executable, b'--version'])
+    if rc != 0 or not out or "version" not in out:
+        if fail:
+            module.fail_json(msg="'%s --version' run failed! Error: %s" %
+                             (executable, err))
+        return None
+    return out.split("version")[1].strip()
--- a/plugins/modules/podman_secret.py
+++ b/plugins/modules/podman_secret.py
@@ -61,6 +61,15 @@ options:
     choices:
       - absent
       - present
+  labels:
+    description:
+      - Labels to set on the secret.
+    type: dict
+  debug:
+    description:
+      - Enable debug mode for module.
+    type: bool
+    default: False
 '''
 
 EXAMPLES = r"""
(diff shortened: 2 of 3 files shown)


plugins/module_utils/podman/podman_container_lib.py
```diff
@@ -53,6 +53,7 @@
     entrypoint=dict(type='str'),
     env=dict(type='dict'),
     env_file=dict(type='path'),
```
sshnaidm, line 55: For example here to change "path" to list of paths

plugins/module_utils/podman/podman_container_lib.py
```diff
@@ -370,6 +371,11 @@ def addparam_env(self, c):
     def addparam_env_file(self, c):
         return c + ['--env-file', self.params['env_file']]
```
sshnaidm, line 372: this function to remove and always treat env_file as a list

plugins/module_utils/podman/podman_container_lib.py
```diff
@@ -53,6 +53,7 @@
     entrypoint=dict(type='str'),
     env=dict(type='dict'),
     env_file=dict(type='path'),
+    env_files=dict(type='list', elements='path'),
```
sshnaidm, line 56: this to make an alias to `env_file`

plugins/module_utils/podman/podman_container_lib.py
```diff
@@ -370,6 +371,11 @@ def addparam_env(self, c):
     def addparam_env_file(self, c):
         return c + ['--env-file', self.params['env_file']]
 
+    def addparam_env_files(self, c):
+        for env_file in self.params['env_files']:
+            c += ['--env-file', env_file]
+        return c
+
```
sshnaidm, line 378: and this to rename to `addparam_env_files`

sshnaidm (commented):
would it work?

sshnaidm:
@ant-bl please rebase on master, fixed CI in #695 
Just a thought - can we maybe just convert `env_file` to list (with alias to env_files) and use it as usual?

ant-bl:
> would it work?

When I tested it works fine

> @ant-bl please rebase on master, fixed CI in #695 

Done

sshnaidm:
@ant-bl  Sorry, probably I didn't explain well. My idea was not to add additional `env_files` argument, but just reuse existing `env_file` parameter making it as list, not as `path` type:
```python
env_file=dict(type='list', elements='path'),
```
```yaml
  env_file:
    description:
      -...
    type: list
    elements: path
    alias:
      - env_files
```
```python
    def addparam_env_file(self, c):
        for env_file in self.params['env_files']:
            c += ['--env-file', env_file]
        return c
```
It will be still backward compatible and will allow to specify multiple env files now.
WDYT?

ant-bl:
> @ant-bl Sorry, probably I didn't explain well. My idea was not to add additional `env_files` argument, but just reuse existing `env_file` parameter making it as list, not as `path` type:
> 
> ```python
> env_file=dict(type='list', elements='path'),
> ```
> 
> ```yaml
>   env_file:
>     description:
>       -...
>     type: list
>     elements: path
>     alias:
>       - env_files
> ```
> 
> ```python
>     def addparam_env_file(self, c):
>         for env_file in self.params['env_files']:
>             c += ['--env-file', env_file]
>         return c
> ```
> 
> It will be still backward compatible and will allow to specify multiple env files now. WDYT?

@sshnaidm Hello. For me it is OK. IMHO the only think against your solution  is that it will break the compatibility of the code. If it is not an issue, it is OK.

PR #697, opened 2023-12-28
Adding --log-opt: max_file to specify the max number of log files. 

AhmadKasem:


plugins/module_utils/podman/podman_container_lib.py
```diff
@@ -473,6 +474,16 @@ def addparam_log_opt(self, c):
                                           errors='surrogate_or_strict')])]
         return c
 
+    def addparam_log_opt_file_size(self, c):
```
sshnaidm, line 477: I don't  think it will work with a separate function, all log options we add in a function above - `addparam_log_opt`.
You can follow same logic with `max_size` and add it like this for example:

```python
    def addparam_log_opt(self, c):
        for k, v in self.params['log_opt'].items():
            if v is not None:
                if k in ('max_file', 'max_size'):
                    opt = k.replace('_', '-')
                c += ['--log-opt',
                      b"=".join([to_bytes(opt, errors='surrogate_or_strict'),
                                 to_bytes(v,
                                          errors='surrogate_or_strict')])]
        return c
```

tests/integration/targets/podman_play/tasks/play-with-options.yml
```diff
@@ -19,6 +19,7 @@
       log_opt:
         path: "{{ log_opt.path }}"
         max_size: "{{ log_opt.size }}"
+        max_file: "{{ log_opt.file }}"
```
sshnaidm, line 22: Please add also check for it in line 40
```yaml
- name: Check annotations and log options are set in Pod
    assert:
      that:
        - play_container_info.containers.0.Config.Annotations["greeting"] == "hello"
        - play_container_info.containers.0.Config.Annotations["greet_to"] == "world"
        - play_container_info.containers.0.HostConfig.LogConfig["Path"] == log_opt.path
        - play_container_info.containers.0.HostConfig.LogConfig["Size"] | lower == log_opt.size

```

sshnaidm (changes requested):

sshnaidm:
@AhmadKasem I've checked now docs and I don't see such option `max-file`: https://docs.podman.io/en/latest/markdown/podman-kube-play.1.html#log-opt-name-value
Where did you see it?

AhmadKasem:
@sshnaidm This is not in the podman docs (not sure why), but it's an option in docker, I just tried using it with podman, and it worked just fine. I also tested the updated ansible module and it worked. 
[Docker documentation for max-file](https://docker-docs.uclv.cu/config/containers/logging/json-file/)

sshnaidm:
> @sshnaidm This is not in the podman docs (not sure why), but it's an option in docker, I just tried using it with podman, and it worked just fine. I also tested the updated ansible module and it worked. [Docker documentation for max-file](https://docker-docs.uclv.cu/config/containers/logging/json-file/)

How did you test it with Podman? Can you reproduce?
I don't see it even in Podman code.

sshnaidm:
Added a feature request in Podman, but till then it's not working: https://github.com/containers/podman/issues/21361
Closing for now

Issue #699, opened 2024-01-10
podman_network not idempotent when disable_dns is explicitly set


4ndyZ:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**
/kind bug

**Description**
Creating a network which explicitly sets `disable_dns: true` other than type `bridge` results in a not idempotent behavior.

This bug was introduced by pull request #650 that fixed #647.

I will create a pull requests with a possible fix later.

**Steps to reproduce the issue:**

1. Configure a Podman network from a different type than `bridge`and explicitly set `dns_enabled`
2. Run the task again.

**Describe the results you received:**
Second run returns `changed` for the task and tries to recreate the network.

**Describe the results you expected:**
Second run should not return `changend` and not recreate the network.

**Additional information you deem important (e.g. issue happens only occasionally):**
The problem is caused by a bug in the logic to compare the problem.

The logic introduced with https://github.com/containers/ansible-podman-collections/pull/650/commits/0551ab46541fa67f923974f37747595181303a47#diff-7cb9aa8fd2e0d6e2319e57721ecfa1b206ef9b56ad9507a6c8df1efac6ab0e1dR334-R338 causes to only get the current configured value when the network type is `bridge`.

This causes logic fails when `dns_enabled` is set to `false` and and is also disabled on the server.

**Version of the `containers.podman` collection:**
```
containers.podman 1.11.0
```

**Output of `ansible --version`:**
```
ansible [core 2.12.10]
  config file = ~/ansible/ansible.cfg
  configured module search path = ['~/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = ~/ansible/lib/python3.11/site-packages/ansible
  ansible collection location = ~/ansible/collections
  executable location = ~/ansible/bin/ansible
  python version = 3.11.7 (main, Dec  4 2023, 18:10:11) [Clang 15.0.0 (clang-1500.1.0.2.5)]
  jinja version = 3.0.3
  libyaml = True
```

**Output of `podman version`:**
```
containers.podman 1.11.0 
```

**Output of `podman info --debug`:**
``` yaml
host:
  arch: amd64
  buildahVersion: 1.31.3
  cgroupControllers:
  - cpuset
  - cpu
  - cpuacct
  - blkio
  - memory
  - devices
  - freezer
  - net_cls
  - perf_event
  - net_prio
  - hugetlb
  - pids
  - rdma
  cgroupManager: systemd
[...]

Issue #647, opened 2023-09-25
podman_network not idempotent with macvlan


kristvanbesien:
/kind bug

**Description**

Creating a macvlan network is not idempotent, unless one explicitly adds " disable_dns: true"

**Steps to reproduce the issue:**

1. Create a mac vlan network in a playbook using the podman_netwokr resource

2. Run playbook a second time. 

**Describe the results you received:**

Second run returns "changed" for the task. Network is recreated. Dependent containers are recreated.

**Describe the results you expected:**

Second playbook run should return "ok"

It looks like this is caused by disable_dns being true by default for macvlan networks, whereas in this module it is false by default. The default should be true except for networks of type bridge. 

**Version of the `containers.podman` collection:**

```
containers.podman             1.10.3
```

**Output of `ansible --version`:**

```
ansible [core 2.15.3]
  config file = /Users/krist/.ansible.cfg
  configured module search path = ['/Users/krist/.ansible/plugins/modules', '/usr/share/ansible/plugins/modules']
  ansible python module location = /Library/Frameworks/Python.framework/Versions/3.11/lib/python3.11/site-packages/ansible
  ansible collection location = /Users/krist/.ansible/collections:/usr/share/ansible/collections
  executable location = /Library/Frameworks/Python.framework/Versions/3.11/bin/ansible
  python version = 3.11.4 (v3.11.4:d2340ef257, Jun  6 2023, 19:15:51) [Clang 13.0.0 (clang-1300.0.29.30)] (/Library/Frameworks/Python.framework/Versions/3.11/bin/python3.11)
  jinja version = 3.1.2
  libyaml = True
```

**Output of `podman version`:**

```
Client:       Podman Engine
Version:      4.6.2
API Version:  4.6.2
Go Version:   go1.20.7
Built:        Mon Aug 28 21:39:48 2023
OS/Arch:      linux/amd64
```

**Output of `podman info --debug`:**

``` yamlhost:
  arch: amd64
  buildahVersion: 1.31.2
  cgroupControllers:
  - cpuset
  - cpu
  - io
  - memory
  - hugetlb
  - pids
  - rdma
  - misc
  cgroupManager: systemd
  cgroupVersion: v2
  conmon:
    package: conmon-2.1.7-2.fc38.x86_64
    path: /usr/bin/conmon
    version: 'conmon version 2.1.7, commit: '
  cpuUtilization:
    idlePercent: 98.56
    systemPercent: 0.48
    userPercent: 0.97
[...]

PR #700, opened 2024-01-10
Fix idempotency for podman_network

4ndyZ:
This pull requests fixes a bug in the diff function (`diffparam_disable_dns`) for the parameter in the `podman_network` plugin (see #699).
It fixes the idempotency for `podman_network` when explicitly definining `disable_dns` parameter for Podman driver other than type `bridge`.

The problem was introduced in #650 (relating issue #647).


sshnaidm:
Thanks for contribution!
Would you like to add maybe tests here? We don't have any non-bridge network test so far..
https://github.com/sshnaidm/ansible-podman-collections/blob/newdiffway1/tests/integration/targets/podman_network/tasks/main.yml

PR #702, opened 2024-01-12
Update env_file to accept a list of files instead of a single file

ant-bl:
I have updated the env_file parameter to use several file to set environment variables.

See https://github.com/containers/ansible-podman-collections/pull/694 for discussion history


plugins/module_utils/podman/podman_container_lib.py
```diff
@@ -369,7 +369,9 @@ def addparam_env(self, c):
         return c
 
     def addparam_env_file(self, c):
-        return c + ['--env-file', self.params['env_file']]
+        for env_file in self.params['env_files']:
```
sshnaidm, line 372:
```suggestion
        for env_file in self.params['env_file']:
```
It should be the real arg, not an alias

sshnaidm (changes requested):

sshnaidm (approved):

sshnaidm:
Thanks for contribution!

PR #703, opened 2024-01-17
Add new arguments to podman status commands

sshnaidm:
Add options to different statuses like start, stop, restart etc.
Refine detach and attach logic.


plugins/module_utils/podman/podman_container_lib.py
```diff
@@ -231,12 +238,35 @@ def construct_command_from_params(self):
 
     def start_stop_delete(self):
 
+        def complete_params(cmd):
+            if self.params['attach'] and self.action == 'start':
+                cmd.append('--attach')
+            if self.params['detach'] is False and self.action == 'start' and '--attach' not in cmd:
```
rabi, line 244: OK, this if block will fix our issue.

rabi (commented):
OK, this looks good to me.

PR #708, opened 2024-02-03
Fix broken example for podman_generate_systemd

dkcampbell:
Change the postgres_local example to use the scope of "user" so the unit file saved to ~/.config/systemd/user/ will be used by systemd.


sshnaidm:
Thanks for fix!

PR #720, opened 2024-02-24
Remove unnecessary quotes in podman_container_exec module

### LINKED ISSUE #714: Env in container exec are wrongly double quoted

/kind bug

**Description**

ENV being forcibly double-quoted, it make the module impossible to use with process that require ENV with strict syntax.

The line inquestion is: https://github.com/containers/ansible-podman-collections/blob/efbfba7c3c4ed95bb75fcabfced61f650b28bac8/plugins/modules/podman_container_exec.py#L159

A possible way would be to add an option to avoid adding these double-quotes around the value and set it to false for retrocompatibility.

**Steps to reproduce the issue:**

For instance, trying to connect to a hashicorp Vault with VAULT_ADDR will fail with the golang error: 

    first path segment in URL cannot contain colon

sshnaidm:
Fix #714
Seems like shell is handling this fine without quotes.

### DIFF
--- a/plugins/modules/podman_container_exec.py
+++ b/plugins/modules/podman_container_exec.py
@@ -162,7 +162,7 @@ def run_container_exec(module: AnsibleModule) -> dict:
 
             to_text(value, errors='surrogate_or_strict')
             exec_options += ['--env',
-                             '%s="%s"' % (key, value)]
+                             '%s=%s' % (key, value)]
 
     if privileged:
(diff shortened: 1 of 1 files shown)


sshnaidm:
@nishipy can you review please?

nishipy:
Thanks for that fix. LGTM.

nishipy:
@sshnaidm LGTM.

PR #722, opened 2024-02-29
Add Quadlet support for Podman

### LINKED ISSUE #671: Add new modules/options for Quadlet

/kind feature

**Description**

Since Podman 4.7.0, the [`podman generate systemd` command has been deprecated](https://github.com/containers/podman/releases/tag/v4.7.0). Instead of that, it is recommended to use `Quadlet` for running containers and pods under systemd.
Now, it's time to consider adding a new module to support Quadlet to run Podman containers under systemd. 

More specifically, [what we need](https://github.com/containers/ansible-podman-collections/issues/671#issuecomment-1844980619) is:
1. Generation of podman Quadlet files in podman_container (and others) as we have today with generate_systemd.
2. Allow podman_container to generate files only, without creating actually containers. (Need to look how easy to do it with current implementation)

Under question:

3. To check the option to reload systemd in the module as it's suggested here: https://github.com/containers/ansible-podman-collections/pull/585
4. To have podman_generate_quadlet module

sshnaidm:
Fix #671


markstos:
I want comment just on the "quadlet" naming here. 

Quadlet made sense as a code name as an external project, but now that it merged into Podman and it is the recommended way to integrate with systemd, the branding is counterproductive unnecessary to expose at the Ansible level. 

The real beauty of of Quadlet is that there is hardly anything new to learn at all. If you know Podman and Systemd, there aren't really new concepts to learn.  Yet Quadlet sounds like yet-another concept that needs to be learned to use Podman and Systemd together. 

The name "Quadlet" appears to be mash-up of the word Quad, meaning "four" and "let" meaning "small", like a piglet is small pig or a parklet is small park, yet what it does has nothing to do with a "small four" as far as I can tell. 

It's also confusing because there are no files related to Podman named Quadlet, no directories named quadlet, no directives named Quadlet-- it's just an abstract concept describing systemd integration.  

The Podman networking features aren't called "octolet", they are described as Podman's networking features. There's no reason to add an extra label to these set of features either.

Let's dig into some specifics in this PR. 

This adds a quadlet as  container state:

> 'absent', 'present', 'stopped', 'started', 'created', 'quadlet'

I  presume what this is trying to express is that the state of container is that it's managed by systemd through a file under /etc/containers/systemd?

That's a bit strange, because systemd services can themselves be present/absent/stopped/started as well. If there's not a clearer way to express this, I would support `systemd` as the name here. 

>  quadlet_file_path=dict(type='path'),
>  quadlet_options=dict(type='list', elements='str'),

Suggested:

  - systemd_file_path
  - systemd_options

> quadlet_file_path is required for Quadlet file generation

Suggest: *systemd_file_path is required for systemd file generation*. 

In summary: I think this would be clearer if the term "systemd" was used everywhere that "quadlet" is used.

nishipy:
In my opinion, it looks clear to use the term `quadlet` here. As far as I know, some official docs and release notes use words including Quadlet, e.g. the terms like Podman Quadlet, Quadlet file appear in https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html. Podman also has env variables like `QUADLET_UNIT_DIRS. In addition, `state: systemd` is confusing for me to distinguish from `podman generate-systemd`. Looking into the source code of Podman, we can find a directory or file named quadlet as of now. 
Technically speaking, Quadlet is a [systemd generator](https://www.freedesktop.org/software/systemd/man/latest/systemd.generator.html), and we will create `qudlet unit files` that Quadlet can translate into units with this module. So `state: quadlet_unit` would be more precise, but `state: quadlet` looks good as well.

markstos:
I understand that `.container` files are used to generate other systemd files. But it looks like a duck and quacks like a duck, and the old duck being deprecated, so I suggest it's called a duck.

sshnaidm:
I think we are going to leave it as `quadlet` now with keeping old (but not removed) `systemd` generaton finctionality. When systemd file generation will be deprecated finally and removed from Podman, we can do alias to `systemd` from `quadlet`.

nishipy (commented):
Leave a comment

plugins/modules/podman_container.py
```diff
@@ -721,6 +724,16 @@
       - Publish all exposed ports to random ports on the host interfaces. The
         default is false.
     type: bool
+  quadlet_file_path:
```
nishipy, line 727: Should we allow the user to input any dir where quadlet unit files are created?
I understand that, as one of the systemd generators, quadlet will scan [the specific directories](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#podman-unit-search-path) during boot and when `systemctl daemon-reload`. Podman has the env var `QUADLET_UNIT_DIRS` but this value seems to be [mainly used for CI tests](https://github.com/containers/podman/blob/v4.9.3/cmd/quadlet/main.go#L107-L113). So it might be better to make only these directories configurable, let's say /etc/containers/systemd/ for rootful and $HOME/.config/containers/systemd/
 for rootless.

sshnaidm: I think it's worth to have some default folder, but also to allow to choose one. Since we have multiple possible folders for that, for example 3 for rootless:
```
$XDG_CONFIG_HOME/containers/systemd/ or ~/.config/containers/systemd/
/etc/containers/systemd/users/$(UID)
/etc/containers/systemd/user
```
But I agree we can take one of them as a default and not to require `quadlet_file_path` to be set.

sshnaidm: Actually we still need to have a file name, so probably we'll need this variable to be set anyway.

alorle: I've been testing this for the last few weeks and I have some questions about what you're talking about here:

# Preamble

According to the official documentation, as @nishipy said, there is only a [limited set of directories](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#podman-rootful-unit-search-path) in which systemd will look for Quadlet files:

- For rootfull: `/etc/containers/systemd` and `/usr/share/containers/systemd`
- For rootless: `$XDG_CONFIG_HOME/containers/systemd`, `~/.config/containers/systemd/`, `/etc/containers/systemd/users/$(UID)` and `/etc/containers/systemd/users`

Also, as specified by the official documentation under [Container units [Container]](https://docs.podman.io/en/latest/markdown/podman-systemd.unit.5.html#container-units-container):

> Container units are named with a `.container` extension and contain a `[Container]` section describing the container that is run as a service. [...] However, some options also affect the details of how systemd is set up to run and interact with the container.
>
> By default, the Podman container has the same name as the unit, but with a `systemd-` prefix, i.e. a `$name.container` file creates a `$name.service` unit and a `systemd-$name` Podman container. The `ContainerName` option allows for overriding this default name with a user-provided one.

As the [name of the container is required](https://docs.ansible.com/ansible/latest/collections/containers/podman/podman_container_module.html#parameter-name) in Ansible's `containers.podman.podman_container` module, the default behaivour of Quadlets is not reacheable.

# Propousal

* Add a `quadlet_basepath` property which allow users select in which path to store the Quadlet files. It could have a default value for rootfull and rootless.
* Use the `name` property for file naming, so the filename is `<name>.container`.
* Delete `quadlet_file_path` property as it would be generated: `<quadlet_basepath>/<name>.container`.
* Add a `container_name` property which allow users override default behaivour for container naming.


*NOTE:* this also applies to other unit types (network, volume, etc)

sshnaidm: @alorle yeah, all this I implement in the new version, came to same conclusions, thanks!

sshnaidm:
I left optional `quadlet_file` (mandatory in case of `kube`) so that people can make custom service name.
`quadlet_dir` is by default one of root/non-root directories and also can be set.

sshnaidm:
@nishipy @alorle and all interested parties, please review, play with it and if no objections we'll merge this week.

plugins/modules/podman_image.py
```diff
@@ -175,6 +176,22 @@
             - docker-daemon
             - oci-archive
             - ostree
+    quadlet_dir:
+      description:
+        - Path to the directory to write quadlet file in.
```
nishipy, line 181: I think it would be better if we describe the default paths:
```suggestion
        - Path to the directory to write quadlet file in. By default, it will be set as `/etc/containers/systemd/` for root user, `~/.config/containers/systemd/` for non-root users.
```

nishipy (changes requested):
@sshnaidm Thank you very much for your hard work! Leave a small comment. 
Overall, it looks good to me. I understand that tests for now don't contain steps of `systemctl --daemon-reload` and checking if containers/pods/images/networks/volumes are created successfully, which might be added after further development.

sshnaidm:
Yes, it can be added in followups.

alorle (approved):
Updated to the latest changes and all is working perfectly. I have tested continers, networks and volumes. Thanks for the hard work @sshnaidm

nishipy (approved):

sshnaidm:
Ok, let's run with it.

Issue #723, opened 2024-03-12
podman_login: why are tlsverify certdir mutually exclusive?


dometto:
BUG REPORT

Using the `podman_login` module with the `tlsverify` and `certdir` arguments set results in an error.

**Description**

The `tlsverify` and `certdir` arguments are explicitly defined as mutually exclusive [here](https://github.com/containers/ansible-podman-collections/blob/master/plugins/modules/podman_login.py#L158-L160). I fail to see the reason for this, as it looks as if I can use the two options with podman:

```
$ podman login --tls-verify --cert-dir bla ghcr.io 
Authenticating with existing credentials for ghcr.io
Existing credentials are valid. Already logged in to ghcr.io
```

However, perhaps I'm missing something about the underlying mechanics that's making it necessary to set these options as exclusive?

If not, I would like to request that these options are made not exclusive, as this is currently leading to problems when using this module with [molecule](https://github.com/ansible/molecule/). A Molecule YAML config like this one currently throws errors:

```
platforms:
  - name: ubuntu
    image: ghcr.io/foo/ubuntu:focal
    tlsverify: $TLS
    certdir: $CERTDIR
```

...as the `tlsverify` and `certdir` parameters are set to an empty string (and thus are both *defined*) when the environment variables are empty. Also see: https://github.com/ansible-community/molecule-plugins/issues/248 

**Steps to reproduce the issue:**

Run the following playbook

```yaml
---
- name: Test podman_login
  hosts: all
  gather_facts: false
  tasks:

    - name: Test
      containers.podman.podman_login:
        certdir: ''
        registry: ghcr.io
        tlsverify: false

```

**Describe the results you received:**

```
fatal: [localhost]: FAILED! => {"ansible_facts": {"discovered_interpreter_python": "/opt/homebrew/bin/python3.11"}, "changed": false, "msg": "parameters are mutually exclusive: certdir|tlsverify"}
```

**Describe the results you expected:**

`tlsverify` and `certdir` should be able to be set together.

**Version of the `containers.podman` collection:** 1.12.0

**Output of `podman version`:**

```Client:       Podman Engine
Version:      4.8.3
API Version:  4.8.3
Go Version:   go1.21.5
Git Commit:   85dc30df56566a654700722a4dd190e1b9680ee7
Built:        Wed Jan  3 12:18:44 2024
OS/Arch:      darwin/arm64

Server:       Podman Engine
Version:      4.9.3
API Version:  4.9.3
[...]

PR #725, opened 2024-03-18
Update podman_login.py

dometto:
Make tlsverify and certdir parameters no longer mutually exclusive. Resolves https://github.com/containers/ansible-podman-collections/issues/723


sshnaidm:
Thanks for the fix!

PR #733, opened 2024-04-16
Create podman secret when skip_existing=True and it does not exist

dmage:


sshnaidm:
@dmage  LGTM, but can you please add the test for this case?

Issue #747, opened 2024-05-19
quadlet creates wrong healthcheck command for container


Marwel:
/kind bug

**Description**

https://github.com/containers/ansible-podman-collections/blob/1ffa40c70593c251ba8ea79836579bc356841be6/plugins/module_utils/podman/quadlet.py#L91

HealthCheckCmd is not a valid command for podman/quadlet
as can be seen
https://github.com/containers/podman/blob/34d2d20d46c1ad3cf13432fc481047f1b285ff20/pkg/systemd/quadlet/quadlet.go#L86

**Steps to reproduce the issue:**

1. Try to deploy a container via quadlet with a healthcheck defined

**Describe the results you received:**
quadlet -dryrun then gives the following error
quadlet-generator[12345]: converting "example_app.container": unsupported key 'HealthCheckCmd' in group 'Container' in /etc/containers/systemd/example_app.container

**Describe the results you expected:**

A working quadlet definition with "HealthCmd" instead of "HealthCheckCmd"

sshnaidm:
fixed by #748

PR #748, opened 2024-05-19
fix(#747): set correct HealthCmd

Marwel:


sshnaidm:
Thanks! Can you please sign your commit with `git commit -s`? It's required by this github org.

PR #751, opened 2024-05-21
Add support for port ranges

frnky:
The commit in this PR implements support for port ranges when using podman_container and podman_pod.

I've run the tests locally with podman 5.0.3 and 4.9.4.


sshnaidm:
@frnky  Thanks for this awesome patch! Although I was just working on [whole redesigning of idempotency ](https://github.com/containers/ansible-podman-collections/pull/750)in the module and it was fixed automatically with new way of difference calculation from `CreateCommand`.
If you would like to contribute more, you can take any issue from what we have. Sorry again for this inconvenience.

frnky:
Good work on the redesign! It felt odd having to implement a parser for the options that are passed to podman to  get the diffing functionality. Reusing the CreateCommand array is definitely the better approach.

PR #619, opened 2023-08-08
Fix transports issues in podman_image

sshnaidm:
Fix #218 #205 #264
Signed-off-by: Sagi Shnaidman <sshnaidm@redhat.com>

PR #755, opened 2024-05-22
Reapply "Fix transports issues in podman_image (#619)" (#754)

sshnaidm:
This reverts commit 391e4c54efb13423c815172398ef87b42a3ae60c.

### DIFF
--- a/plugins/modules/podman_image.py
+++ b/plugins/modules/podman_image.py
@@ -176,6 +176,7 @@ DOCUMENTATION = r'''
           type: str
           choices:
             - dir
+            - docker
             - docker-archive
             - docker-daemon
             - oci-archive
@@ -557,16 +558,8 @@ class PodmanImageManager(object):
                 self.results['changed'] = True
 
         if self.push:
-            # Push the image
-            if '/' in self.image_name:
-                push_format_string = 'Pushed image {image_name}'
-            else:
-                push_format_string = 'Pushed image {image_name} to {dest}'
-            self.results['actions'].append(push_format_string.format(image_name=self.image_name, dest=self.push_args['dest']))
-            self.results['changed'] = True
-            if not self.module.check_mode:
-                self.results['image'], output = self.push_image()
-                self.results['stdout'] += "\n" + output
+            self.results['image'], output = self.push_image()
+            self.results['stdout'] += "\n" + output
         if image and not self.results.get('image'):
             self.results['image'] = image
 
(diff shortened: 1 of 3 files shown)


plugins/modules/podman_image.py
```diff
@@ -806,19 +799,20 @@ def push_image(self):
             else:
                 dest_format_string = '{transport}:{dest}'
 
-        dest_string = dest_format_string.format(transport=transport, name=self.name, dest=dest, image_name=self.image_name,)
+            dest_string = dest_format_string.format(transport=transport, name=self.name, dest=dest, image_name=self.image_name,)
+        else:
+            if len(dest.split('/')) < 3:
+                dest_string = f'{dest}/{self.image_name}'
+            else:
+                dest_string = dest
```
cjeanner, line 807:
```suggestion
            if dest and len(dest.split('/')) < 3:
                dest_string = f'{dest}/{self.image_name}'
            elif dest:
               dest_string = dest
            else:
                dest_string = name
```

would cover the case where  `dest` isn't set, but `name` is a valid replacement (this is checked on L784-786)

plugins/modules/podman_image.py
```diff
@@ -781,19 +781,12 @@ def push_image(self):
         # Build the destination argument
         dest = self.push_args.get('dest')
         dest_format_string = '{dest}/{image_name}'
-        regexp = re.compile(r'/{name}(:{tag})?'.format(name=self.name, tag=self.tag))
         if not dest:
             if '/' not in self.name:
                 self.module.fail_json(msg="'push_args['dest']' is required when pushing images that do not have the remote registry in the image name")
```
cjeanner, line 787: Doing so, we can do all the checks in one shot - and remove the `if not dest: fail` on L793-794
```suggestion
        if not dest:
            if transport:
                self.module.fail_json(msg="'ppush_args['transport'] requires 'push_args['dest'] but it was not provided.")
            if '/' not in self.name:
                self.module.fail_json(msg="'push_args['dest']' is required when pushing images that do not have the remote registry in the image name")
                
```
It's not much, but it  would allow to do all the "dest_string" building afterward, in a far more convenient way.
We could also apply a plain/stupid:
```YAML
if not dest:
  dest = self.name
```

so that we can stick to `dest` for all the rest of the code.

BTW: `transport` wouldn't work with a `name` matching a `dest` syntax?

tests/integration/targets/podman_image/tasks/idem_push.yml
```diff
@@ -0,0 +1,445 @@
...
+  ignore_errors: true
+  register: img_result1
+
+- name: Check outputs
+  assert:
+    that:
+      - img_result1 is failed
+      - "'Destination must be a full URL or path to a directory' in img_result1.msg"
+
+- containers.podman.podman_image:
```
cjeanner, line 15: care to add another test with `name: quay.io/testing/testimage:my-tag`? That's what we're using on our side, ensuring this is working as well and won't break in the future would be nice :). I don't think you covered that use-case in the tests.
It would probably exist like img_result2 (auth required) with "my-tag" instead of "latest". This is to ensure the `:TAG` is properly detected.

PR #784, opened 2024-06-12
params gpus should be exit_policy

richm:
The code should be setting `--exit-policy` to the value of `params['exit_policy']` instead of `params['gpus']`


richm:
@sshnaidm this was caught by internal static analysis

sshnaidm:
> @sshnaidm this was caught by internal static analysis

Nice catch! What do you use for internal static analysis?

PR #809, opened 2024-07-31
Fix health-startup-cmd

HolgerHees:
"--health-startup-command" is a wrong parameter.

the correct syntax is

"--health-startup-cmd"


sshnaidm:
Great, thanks for catching this!

PR #835, opened 2024-08-29
Fix quadlet parameters when container uses rootfs

mainnika:
Adjust quadlet parameters to set rootfs as image and clear image field when rootfs container is being used.

e.g.:
```
    rootfs: true
    image: "/opt/container-root"
```

should put into a quadlet file (note, no `Image=` in a file):
```
Rootfs=/opt/container-root
```


sshnaidm:
Great, thanks for contribution @mainnika !

Issue #830, opened 2024-08-28
push_args destination now requires full path with image name and tag when default transport is used


jakestec:
Hello,

I've been using the latest collection version (v1.15.4) with an existing playbook to build container images. When I went to push a new image today I got the following error related to the line of code below:

`Destination must be a full URL or path to a directory.`

https://github.com/containers/ansible-podman-collections/blob/552af3ef7ff62706ec4bd88e8f962e069c66eddc/plugins/modules/podman_image.py#L795

My task:
```
container_registry: "my.registry.domain.name"
...
    - name: Create and push EE => "{{ ee_name }}:{{ ee_tag }}"
      containers.podman.podman_image:
        name: "{{ ee_name }}"
        tag: "{{ ee_tag }}"
        path: "{{ ansible_env.PWD }}/build"
        push: true
        auth_file: "{{ lookup('ansible.builtin.env', 'XDG_RUNTIME_DIR') }}/containers/auth.json"
        validate_certs: true
        push_args:
          dest: "{{ container_registry }}"
```
Debugging I can see its not forming the push arguments correctly:
```
    "podman_actions": [
        "/usr/bin/podman image ls rcnet-ee:1.9 --format json",
        "/usr/bin/podman inspect rcnet-ee:1.9 --format json",
        "/usr/bin/podman push --tls-verify --authfile /run/user/1001/containers/auth.json my-ee:1.9 my.registry.domain.name",
```
We should see:
```
 "/usr/bin/podman push --tls-verify --authfile /run/user/1001/containers/auth.json my-ee:1.9 my.registry.domain.name/my-ee:1.9",
```

The above task is working in v1.12.1. I was able to fix the issue by modifying the destination in the task to include the image name and tag like so:

```
    - name: Create and push EE => "{{ ee_name }}:{{ ee_tag }}"
      containers.podman.podman_image:
        name: "{{ ee_name }}"
        tag: "{{ ee_tag }}"
        path: "{{ ansible_env.PWD }}/build"
        push: true
        auth_file: "{{ lookup('ansible.builtin.env', 'XDG_RUNTIME_DIR') }}/containers/auth.json"
        validate_certs: true
        push_args:
          dest: "{{ container_registry }}/{{ ee_name }}:{{ ee_tag }}"
```

[...]

PR #848, opened 2024-09-23
Update error and docs for new dest functionality in podman image

jakestec:
Updating documentation and the error related to setting destination in the podman_image task to be in line with new functionality.

Discussion here: https://github.com/containers/ansible-podman-collections/issues/830

### DIFF
--- a/plugins/modules/podman_image.py
+++ b/plugins/modules/podman_image.py
@@ -794,7 +794,7 @@ class PodmanImageManager(object):
                 dest_string = dest.rstrip("/") + "/" + self.image_name
 
         if "/" not in dest_string and "@" not in dest_string and "docker-daemon" not in dest_string:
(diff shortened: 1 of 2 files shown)


plugins/modules/podman_image.py
```diff
@@ -794,7 +794,7 @@ def push_image(self):
                 dest_string = dest.rstrip("/") + "/" + self.image_name
 
         if "/" not in dest_string and "@" not in dest_string and "docker-daemon" not in dest_string:
-            self.module.fail_json(msg="Destination must be a full URL or path to a directory.")
+            self.module.fail_json(msg="Destination must be a full URL or path to a directory with image name and tag.")
```
sshnaidm, line 797:
```suggestion
            self.module.fail_json(msg="Destination must be a full URL with image name and tag or path to a directory.")
```

sshnaidm:
Thanks!

PR #852, opened 2024-09-25
don't document quadlet_dir as required when setting state=quadlet

evgeni:
The underlying code has sane defaults and will figure out the directory on its own in most cases
https://github.com/containers/ansible-podman-collections/blob/0b773e5d61d1f0c8ea33b8f42dafe2267ee8e487/plugins/module_utils/podman/quadlet.py#L688-L693


sshnaidm:
Thanks for contribution!

PR #860, opened 2024-10-08
podman_login does not support check_mode

mstinsky:
The podman_login module does not have code to support check runs. 
Therefore a check run in ansible does the actual login. 
Set supports_check_mode to False so the login is skipped in a dry-run.


sshnaidm:
Thanks for the fix!

PR #867, opened 2024-10-14
Add quadlet file mode option to specify file permission

### LINKED ISSUE #864: Add option to define quadlet file permission
/kind feature

**Description**

Quadlet files are created with the `0644` permission, which allows all users to read the quadlet file.
This can lead to security issues. For example, if sensitive information is defined in environment variables, it will be visible to all users.
Currently, an additional task is required to set more restrictive permissions on the generated files.

As a user, it would be nice to have an option to specify the mode of the quadlet file, for example `quadlet_file_m̀ode`.
And I suggest setting it to `0640` by default.

ghoudmon:
Fixes #864


sshnaidm:
Thanks for the contribution!

PR #882, opened 2024-12-16
Fix podman_container_copy examples

derhagen:
The podman_container_copy examples use the podman_search module


sshnaidm:
Please sign your commit with `git commit -s`, thanks for update!

Issue #892, opened 2025-01-18
Creating a container with an environment variable containg "MAX_SIZE" is not idempotent


Popkornium18:
Example:

```
---
- hosts: localhost
  tasks:
    - containers.podman.podman_container:
        name: nginx
        image: docker.io/nginx:latest
        env:
          THIS_IS_THE_MAX_SIZE: 9001
```

When this is executed with `--check --diff` we can see this output:

```
--- before
+++ after
@@ -1 +1 @@
-env - this_is_the_max_size=9001
+env - this_is_the_max-size=9001
```

This issue was introduced in https://github.com/containers/ansible-podman-collections/pull/862

PR #893, opened 2025-01-18
Fix idempotency for containers with env vars containing MAX_SIZE

Popkornium18:
Fixes https://github.com/containers/ansible-podman-collections/issues/892, a regression introduced in https://github.com/containers/ansible-podman-collections/pull/862.


sshnaidm (approved):

sshnaidm:
Thanks for fixing this!

Issue #902, opened 2025-02-14
podman_image is incompatible with python <3.9


sents:
The keyword argument `usedforsecurity` of the hashlib.sha256 function was [introduced in python3.9](https://docs.python.org/3.12/library/hashlib.html#hashlib-usedforsecurity).
https://github.com/containers/ansible-podman-collections/blob/010410ae316700bbd50ee64c13e823880969c5a5/plugins/modules/podman_image.py#L560

This means that the podman_image is not compatible with python versions older than 3.9.

As current versions of OpenSuse Leap and Suse Enterprise Linux are unfortunately still using python 3.6 this makes the
podman_image module unavailable on these distributions.

sents:
Upon reading about `usedforsecurity` I am wondering if it is really applicable here. It forces use of secure hash functions in specially compiled python versions, upgrading insecure algorithms in case. In this case we are comparing with podman image hashes which are sha256 so an upgraded different algorithm would never match an existing build image. 

This can be understood as being more secure in case we don't trust the host images. But in this case we should maybe also hash the contents in the directory of the Container file if it contains any COPYs.

sshnaidm:
Thanks for pointing this out, I'd like to keep it available for 3.6 of course. I think we can strip this keyword or to include it only when we run on > 3.9
Would you like to send a patch?

sents:
Fixed in #904

PR #904, opened 2025-02-17
Use usedforsecurity for hashlib.sha256 only in python version >=3.9

sents:
The usedforsecurity keyword argument of the hashlib functions was introduced in python 3.9. To achieve compatibility with versions below that, we only use it once it is available.

The usedforsecurity argument forces use of secure hash functions in specially compiled versions of python. In this case it would force to upgrade sha256 to a different hash function should sha256 be deemeed insecure in the future. The podman hash we are comparing against is (currently) always sha256.

As sha256 is still considered secure, removing this option for older python versions should be acceptable.

Related Issues: #902

### DIFF
--- a/plugins/modules/podman_image.py
+++ b/plugins/modules/podman_image.py
@@ -432,6 +432,7 @@ import shlex  # noqa: E402
 import tempfile  # noqa: E402
 import time  # noqa: E402
 import hashlib  # noqa: E402
+import sys  # noqa: E402
 
 from ansible.module_utils._text import to_native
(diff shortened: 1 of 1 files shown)


sshnaidm:
Awesome, please sign your commit with `git commit -s`

sshnaidm:
Need to fix linters:
`ERROR: plugins/modules/podman_image.py:435:11: E261: at least two spaces before inline comment`

sents:
Great, thank you :slightly_smiling_face:

PR #911, opened 2025-03-07
Document that sdnotify can be set to healthy

ekohl:
For quadlets you can set sdnotify (which maps to Notify=) to healthy to use a healthcheck to determine when the container is up.


sshnaidm:
Thanks, please sign your commit with `git commit -s`

sshnaidm:
Thanks!

PR #915, opened 2025-03-18
Fix config precedence and config source

bcoca:
`inventory_hostname` should have lowest precedence

`play_context` can have stale data, it is kept for backwards compatibility and fallback (for very old Ansible core version). Configuration should be pulled primarily from `get_option`.

### DIFF
--- a/plugins/connection/podman.py
+++ b/plugins/connection/podman.py
@@ -22,8 +22,8 @@ DOCUMENTATION = '''
           - The ID of the container you want to access.
         default: inventory_hostname
         vars:
-          - name: ansible_host
           - name: inventory_hostname
+          - name: ansible_host
           - name: ansible_podman_host
       remote_user:
         description:
@@ -87,11 +87,11 @@ class Connection(ConnectionBase):
     def __init__(self, play_context, new_stdin, *args, **kwargs):
         super(Connection, self).__init__(play_context, new_stdin, *args, **kwargs)
 
-        self._container_id = self._play_context.remote_addr
+        self._container_id = self.get_option('remote_addr') or self._play_context.remote_addr
         self._connected = False
         # container filesystem will be mounted here on host
         self._mount_point = None
-        self.user = self._play_context.remote_user
+        self.user = self.get_option('remote_user') or self._play_context.remote_user
         display.vvvv("Using podman connection from collection")
 
     def _podman(self, cmd, cmd_args=None, in_data=None, use_container_id=True):


sshnaidm:
Thanks, please sign your commit with `git commit -s`.

sshnaidm:
For Ansible < 2.10 it fails with:
```
An exception occurred during task execution. To see the full traceback, use -vvv. The error was: AttributeError: 'Connection' object has no attribute '_load_name'
```
for newer version:
```
TASK [raw with unicode arg and output] *****************************************
fatal: [podman-container]: FAILED! => {"changed": true, "msg": "non-zero return code", "rc": 125, "stderr": "Error: no container with name or ID \"inventory_hostname\" found: no such container\n", "stderr_lines": ["Error: no container with name or ID \"inventory_hostname\" found: no such container"], "stdout": "", "stdout_lines": []}
```

bcoca:
> Thanks, please sign your commit with `git commit -s`.

I used the web interface, to do this i would have to redo patch as a checkout

sshnaidm:
> 2.10 is ancient and not supported, in any case `_load_name` is something the core plugin loader attaches to the plugin itself on load, not something that the plugin should supply. This error seems like a red herring.

This is actually for 2.9, not for 2.10 (**Ansible < 2.10**). I know it's still used and if we want to break all 2.9 users it should be a major release which breaks backward compatibility.

> 
> this seems like a templating issue that was fixed in core a while back. I did not see such error when using podman with 2.18 (current stable), I'll check the specific versions later when I get a chance.

This happened in jobs with Ansible `2.17` and `core 2.19.0.dev0` from `devel` branch.

> I used the web interface, to do this i would have to redo patch as a checkout

This is a requirement for all patches in `containers` github org, not my nit.

bcoca:
Sorry about the noise, I just upgraded my homelab to podman from lxd setup (i know, ancient). In the process I used your connection plugin to take over what i did with the lxc/lxd connections and saw the issues with the incorrect priority in config, fixed locally and then when I came back from PTO I remembered the issues and tried to share fix here.  I did not mean to bypass your acceptance rules, I was just explaining why/how i created the PR and why it was not signed and might need to be closed in favor of new one.

The 2.9 support really surprised me, but that will make any PR I create incompatible as 2.9 has the bug that won't template `inventory_hostname` when it appears as config variable, but the other variables will also have issues as `play_context` will not recognize the custom one for the plugin (issue that took me a good while to debug!).

If you are open to accept a revised PR, it would limit the collection to >=2.14 and above (which contains the fix to `inventory_hostname` templating) .. I'll also look into the other issues CI encountered (I might have not created correct patch from memory as I did at home).

sshnaidm:
@bcoca No worries, I'll take a look on next week on what's wrong there. It should be fixed at least for latest Ansible versions, not even 2.9.
This is very old code and probably mostly irrelevant now, if I had time to work on it I'd rewrite it from scratch. :worried:

Issue #807, opened 2024-07-18
Quadlet Label Value Truncation Due to Missing Quotes


tjmullicani:
**Is this a BUG REPORT or FEATURE REQUEST? (leave only one on its own line)**

/kind bug

**Description**
When a Quadlet file is written, the "Label=" line values are not quoted. This causes issues if there are special or escape characters in the value.

For example, the following quadlet file is currently generated.

```text
[Container]
ContainerName=testapp
...
Label=traefik.enable=True
Label=traefik.http.routers.testapp.rule=( Host(`test.com`) && PathPrefix(`/path`) )
```

When it should be (notice the extra quotes around `Label=`.

```text
[Container]
ContainerName=testapp
...
Label=traefik.enable="True"
Label=traefik.http.routers.testapp.rule="( Host(`test.com`) && PathPrefix(`/path`) )"
```

**Steps to reproduce the issue:**

```yaml
- containers.podman.podman_container:
     name: testapp
     state: quadlet
     quadlet_filename: container-testapp
     quadlet_options:
       - "AutoUpdate=registry"
       - "Pull=always"
       - |
         [Install]
         WantedBy=default.target
    network:
      - podman
    image: docker.io/containous/whoami
    labels:
      traefik.enable: true
      traefik.http.routers.testapp.rule: "( Host(`test.com`) && PathPrefix(`/path`) )"
```

**Describe the results you received:**

```shell-session
$ podman inspect testapp
...
      "Labels": {
        "traefik.http.routers.testapp.rule": "(",
      }
```

**Describe the results you expected:**

```shell-session
$ podman inspect testapp
...
      "Labels": {
        "traefik.http.routers.testapp.rule": "( Host(`test.com`) && PathPrefix(`/path`) )",
      }
```

Double quoting the label within containers.podman.podmain_container is a workaround.

```yaml
...
- containers.podman.podman_container:
    ...
    labels:
      traefik.http.routers.testapp.rule: "\"( Host(`test.com`) && PathPrefix(`/path`)\""
...
```

**Version of the `containers.podman` collection:**
**Either git commit if installed from git: `git show --summary`**
**Or version from `ansible-galaxy` if installed from galaxy: `ansible-galaxy collection list | grep containers.podman`**

```text
1.15.4
```

**Output of `ansible --version`:**

```text
ansible [core 2.15.10]
```

**Output of `podman version`:**

```shell-session
$ podman --version
podman version 4.9.4-rhel
```

[...]

PR #920, opened 2025-04-02
correctly quote labels and environment variables for quadlets

evgeni:
Fixes: #807


sshnaidm:
LGTM, thanks for fixing this!

PR #929, opened 2025-05-04
Fix idempotency for any podman secret driver

lersveen:
This adds proper idempotency for secret creation in module `podman_secret` regardless of the driver used.

As I understand it, all secret drivers are provided with the exact same interface in Podman (at least from v4.7.0?), so there should be no need to hardcode the state as `changed` for all drivers other than 'file'.

### DIFF
--- a/plugins/modules/podman_secret.py
+++ b/plugins/modules/podman_secret.py
@@ -140,11 +140,6 @@ def need_update(module, executable, name, data, path, env, skip, driver, driver_
         return False
     try:
         secret = module.from_json(out)[0]
-        # We support only file driver for now
-        if (driver and driver != 'file') or secret['Spec']['Driver']['Name'] != 'file':
-            if debug:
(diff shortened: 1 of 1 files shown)


sshnaidm:
@lersveen thanks for the contribution! LGTM, but can you please add a test with other driver? For example here: https://github.com/containers/ansible-podman-collections/blob/main/tests/integration/targets/podman_secret/tasks/main.yml#L284

lersveen:
Wrote some tests, though I'm unable to immediately figure out how to run them locally?

sshnaidm:
@lersveen I just include the tasks from that path and run locally.
```yaml
- hosts: all
  tasks:
    - include_tasks: ....
```
just make sure you use module with your changes

sshnaidm:
@lersveen all works like a charm, thanks for your contribution! :clap:

Issue #940, opened 2025-06-20
podman_container fails to identify changes on the volume mount opts


lucas-benedito:
When using the `containers.podman.podman_container` module, the `volume` parameter only identifies changes on the path; however, it does not identify changes on the mount options.

E.g.
Before:
```
  containers.podman.podman_container:
      ...
      volume: 'redis_run:/run/redis:z'
```
After:
```
  containers.podman.podman_container:
      ...
      volume: 'redis_run:/run/redis:U'
```

The module fails to identify the change and subsequently does not recreate the container.

Note: When first creating the container, the volume receives the proper volume mount options and only fails to modify it.

PR #941, opened 2025-06-20
podman_container_lib: Added checks for volume opts

lucas-benedito:
Changed the diffparam_volume function to include the volume mount opts.

Fix: #940


lucas-benedito:
Test added to the suggested file. I had to create a temporary directory, as other file systems like /opt and /tmp would have issues with SELinux while setting the parameters.

sshnaidm:
Awesome! Thanks for contribution!

